PR #345, opened 2014-10-14
Easy ramp up for Node.js by auto-loading CLDR data

rxaviers:
We're able to offer an extra level of convenience for Node.js users by auto-loading necessary CLDR data.

Note this code resides in the special file that is only loaded for Node.js environment. So, it has no implications for usage in any other environment.


rxaviers:
Auto loading CLDR didn't sound as a good idea. Closing.

PR #367, opened 2014-12-16
Build: Add config for browserstack-runner

jzaefferer:
This is a work in progress, to debug the issues we're still having with this setup.

### DIFF
--- a/browserstack-ie8.json
+++ b/browserstack-ie8.json
@@ -1,0 +1,7 @@
+{
+  "test_framework": "qunit",
+  "test_path": [ "test/unit-es5-shim.html", "test/functional-es5-shim.html" ],
+  "browsers": [
+    "ie_8"
+  ]
+}
--- a/browserstack.json
+++ b/browserstack.json
@@ -1,0 +1,17 @@
+{
+  "test_framework": "qunit",
+  "test_path": [ "test/unit.html", "test/functional.html" ],
+  "browsers": [
+    "chrome_previous",
+    "chrome_latest",
+    "firefox_previous",
+    "firefox_latest",
+    "ie_9",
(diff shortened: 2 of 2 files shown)


jzaefferer:
Updated to include both unit and functional tests and a separate config for IE8. Can run the latter with:

```sh
BROWSERSTACK_JSON=browserstack-ie8.json browserstack-runner
```

Both ran successfully on my local machine. I'm still working on the Travis integration.

browserstack-ie8.json
```diff
@@ -0,0 +1,7 @@
+{
+  "test_framework": "qunit",
+  "test_path": ["test/unit-es5-shim.html", "test/functional-es5-shim.html"],
```
rxaviers: Space `[ "test/unit-es5-shim.html", "test/functional-es5-shim.html" ]`?

browserstack.json
```diff
@@ -0,0 +1,17 @@
+{
+  "test_framework": "qunit",
+  "test_path": ["test/unit.html", "test/functional.html"],
```
rxaviers: Space `test_path": [ "test/unit.html", "test/functional.html" ],`?

jzaefferer:
Btw. since the Travis integration is still problematic, we can start by only landing the config files, so that we can use them locally. We can land the other changes later.

rxaviers:
Sounds good to me.

rxaviers:
Sounds good to me. Do you want to update this branch accordingly and potentially creating a new PR to track Travis integration?

rxaviers:
> Can run the latter with:
> `BROWSERSTACK_JSON=browserstack-ie8.json browserstack-runner`

This is really cool!

PR #373, opened 2014-12-23
Date: Fix parsing of the day part when it precedes the month part

### LINKED ISSUE #323: parseDate: incorrect result when parsing last day of month
Hi, there seems to be a problem in date/parse.js with parsing a date which day part is greater than amount of days current month has and a day part is parsed before a month part.
Today it's 30.09.2014 and parsing "31.10.2014" would yield 30.10.2014. The reason it happens is that initial date is current date and for September 31.09.2014 is an incorrect date.

mateubo:
Fixes #323

### DIFF
--- a/src/date/parse.js
+++ b/src/date/parse.js
@@ -247,7 +247,7 @@ return function( value, tokens, properties ) {
 		}
 		date.setDate( day );
 	} else if ( totalDays !== undefined ) {
-		if ( outOfRange( totalDays, 1, dateIsLeapYear( date.getFullYear() ) ? 366 : 355 ) ) {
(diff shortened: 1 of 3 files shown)


src/date/parse.js
```diff
@@ -20,7 +21,7 @@ define([
  * ref: http://www.unicode.org/reports/tr35/tr35-dates.html#Date_Format_Patterns
  */
 return function( value, tokens, properties ) {
-	var amPm, era, hour, hour12, timezoneOffset, valid,
+	var amPm, day, era, hour, hour12, timezoneOffset, totalDays, valid,
```
mateubo: Ok, done.

rxaviers:
Thank you very much @mateubo, just let us know if you find any other problem.

Issue #235, opened 2014-04-16
Testing automation


jzaefferer:
We want to test this in all supported browsers automatically. I'll look into creating a configuration file for `browserstack-runner` to see if that works here. Otherwise we can try it with TestSwarm.

rxaviers:
WIP PR #450

rxaviers:
Close in favor of #450

rxaviers:
Actually, reopen. Should be fixed by #450.

PR #374, opened 2014-12-30
Date: Create unit tests to verify day parsing uses the correct day range given its corresponding month/year

### LINKED ISSUE #323: parseDate: incorrect result when parsing last day of month

rxaviers:
Fixes #323
Ref #373

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -14,11 +14,13 @@
   },
   "devDependencies": {
     "es5-shim": "3.4.0",
+    "lolex": "sinonjs/lolex#3664eb",
     "make-plural": "eemeli/make-plural.js#2.1.2",
     "messageformat": "SlexAxton/messageformat.js#debeaf4",
     "qunit": "1.16.0",
     "requirejs": "2.1.9",
     "requirejs-plugins": "1.0.2",
-    "requirejs-text": "2.0.10"
+    "requirejs-text": "2.0.10",
+    "sinon": "1.12.2"
   }
 }
--- a/bower.json
+++ b/bower.json
@@ -13,8 +13,8 @@
     "cldr-data": ">=25"
   },
   "devDependencies": {
-    "make-plural": "eemeli/make-plural.js#2.1.2",
     "es5-shim": "3.4.0",
+    "make-plural": "eemeli/make-plural.js#2.1.2",
     "messageformat": "SlexAxton/messageformat.js#debeaf4",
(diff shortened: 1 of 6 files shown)


jzaefferer:
Specifically the clock feature: http://sinonjs.org/docs/#clock

```js
var clock = sinon.useFakeTimers(now);
```

Where `now` is your custom "today" as a timestamp. That should do the trick.

rxaviers:
Awesome. I've updated this PR to use Sinon.js instead of my custom implementation on tests.

jzaefferer:
Looks good to me.

rxaviers:
Testing this PR+Sinon.js across various browsers resulted in errors/failures (I've merged in the browserstack-runner branch locally to test it). Running browserstack-runner against my custom implementation passes just fine.

- Sinon.js on browserstack: [fail](https://gist.github.com/rxaviers/8f0ad35f478941f2da8d).
- Custom `FakeDate` on browserstack: [pass](https://gist.github.com/rxaviers/c028c0cff270b88061ad).

Potentially, the Sinon.js failures are only related to AMD loading, giving the fact some failing tests intermittently pass.

rxaviers:
We could land this PR + Sinon.js now with no problem. But, it would impact (i.e. delay) testing automation #235 (i.e., to land browserstack-runner branch #367), since we would have to figure out what this failures were.

Giving test automation (which is due to 1.0.0) and #323 have higher priorities compared to using Sinon.js to simplify FakeDate here on tests, I'm considering to track that as a separate issue unless anyone has a different idea. So, we can land #373 (fixing #323), land browserstack-runner branch. Then, spend time on it.

What do you think?

jzaefferer:
Is the sinon/browserstack failure consistent? Can you reproduce it when running the given browser manually?

If we're not getting anywhere, I'm okay with keeping `FakeDate` for now.

rxaviers:
It's not consistent. I haven't tried it manually. The problem could be simple. But, it will require some time investigating. I just thing this time is best spent somewhere else now. :)

jzaefferer:
Okay, let's give sinon another try later.

PR #375, opened 2014-12-31
Add commitplease to validate commit messages via grunt default task

### LINKED ISSUE #347: Add commitplease as devDependency
Globalize currently isn't using commitplease. Is there any problem with it?

rxaviers:
Fixes #347

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -28,6 +28,13 @@ module.exports = function( grunt ) {
 
 	grunt.initConfig({
 		pkg: pkg,
+		commitplease: {
+			currentBranch: {
+				options: {
+					committish: "cf52d58..HEAD"
+				}
+			}
+		},
 		connect: {
 			options: {
 				port: 9001,
@@ -407,7 +414,8 @@ module.exports = function( grunt ) {
 		// "jscs:dist",
 		"test:functional",
 		"uglify",
-		"compare_size"
+		"compare_size",
+		"commitplease"
 	]);
 
 };
--- a/package.json
+++ b/package.json
@@ -72,6 +72,7 @@
   "devDependencies": {
     "grunt": "0.4.5",
     "grunt-check-dependencies": "0.6.0",
+    "grunt-commitplease": "0.0.3",
     "grunt-compare-size": "0.4.0",
     "grunt-contrib-clean": "0.6.0",
     "grunt-contrib-connect": "0.8.0",


jzaefferer:
Is this supposed to land together with #368, or to replace it?

jzaefferer:
Since the task implementation is still in a branch: If you do a PR against your own repo, I can do a code review there.

rxaviers:
Jörn, I lost your comment in my emails, sorry for the delay.

This is supposed to replace #368. I've created a PR there (https://github.com/rxaviers/grunt-commitplease/pull/1).

jzaefferer:
I still don't understand why you don't want to land #368. I've been using commitplease as-is in many of our projects and its very useful to get immediate feedback while committing.

dmethvin:
As a contributor I think it's more useful to get early feedback about problems than waiting for the CI or build process to say something. I don't think Globalize has had this problem yet but their haven't been a lot of contributors.

rxaviers:
> I still don't understand why you don't want to land #368. 

@jzaefferer, according to our conversation on Dec 12, I understood we agreed on landing #368 unless we had this solution (using grunt task).

> As a contributor I think it's more useful to get early feedback about problems than waiting for the CI or build process to say something

@dmethvin, I also think it's important that contributors (+ people in general) have freedom to choose what is the most convenient for them. Everyone can opt in on using the git hook. But, not opt out. Everyone that has a preference for this sort of feedback can use the git hook with no problem. I can add instructions on how to set it up. The other way around is not true (forcing the git hook on everyone inhibits their choice of not wanting this sort of feedback).

rxaviers:
The ultimate goal of #347 is that no incorrect commit message lands on master. Please, correct me if I'm wrong.

The approach implemented on #368 doesn't address that (given our discussion there on #347 - see [[1]](https://github.com/jquery/globalize/issues/347#issuecomment-60112437) and [[2]](https://github.com/jquery/globalize/issues/347#issuecomment-60113254)). In the other hand, the approach implemented here does, because all branches will get a failed CI status if a commit message is wrong.

Please, just let me know on any questions or different ideas.

arschmitz:
@rxaviers unless you plan on not following the commit message guidelines for globalize what would be the point in disabling commitplease as a git hook? A grunt task is nice i guess, but for new contributors ( the ones most likely to run into this ). It will be much harder to fix a message ( or messages ), after finding them with the grunt task, then it would have been to just fix the message, when they made the commit in the first place. Also instructions to set up a git hook are great except that new contributors ( again the ones i think benefit the most from the hook ), would have no idea to do this in the first place.

arschmitz:
@rxaviers Personally I think its about more then just preventing bad messages from landing on master, it is also about helping contributors to find and fix these messages before they even get to the PR stage ( as early as possible ) . Yes there are ways someone can work around the hook, but i don't think we have run into that on any of the many repos using commitplease currently. If you want to be extra sure with a grunt task thats awesome, but it seems like this does not replace the git hook to me.

rxaviers:
@arschmitz, please read [[1]](https://github.com/jquery/globalize/issues/347#issuecomment-60112437) and [[2]](https://github.com/jquery/globalize/issues/347#issuecomment-60113254). New contributors will submit their fixes/changes via a PR. The PR automatically runs CI and we automatically get notified whether or not there's a incorrect commit message.

Note we use the same workflow for everything else: coding standards (via jscs grunt task), tests (via qunit grunt task), lint errors (via jshint grunt task). Why would it be different with commit message standards.

arschmitz:
@rxaviers please re-read my comment what i said was.
> it is also about helping contributors to find and fix these messages before they even get to the PR stage

Key word being "before" a git hook will find these before they ever submit a pr, or even make the commit.

rxaviers:
> it is also about helping contributors to find and fix these messages before they even get to the PR stage

So, we need to prevent them from saving their files if jscs or jshint or test fails.

rxaviers:
> I also think it's important that contributors (+ people in general) have freedom to choose what is the most convenient for them. Everyone can opt in on using the git hook. But, not opt out. Everyone that has a preference for this sort of feedback can use the git hook with no problem. I can add instructions on how to set it up. The other way around is not true (forcing the git hook on everyone inhibits their choice of not wanting this sort of feedback).

@arschmitz, this is an important statement in my opinion. Please, consider that people can opt in using the git hook if that's their choice. But, not to opt out.

arschmitz:
@rxaviers I think thats a bit of a stretch we are only stopping a commit if the COMMIT MESSAGE fails the test. We are not preventing saving, we are simply asking them to retype the message in the proper format.

arschmitz:
@rxaviers as i already asked. 

> unless you plan on not following the commit message guidelines for globalize what would be the point in disabling commitplease as a git hook?

I don't think we should allow opting out what is the use case?

dmethvin:
There isn't really a need for opt-out at the commitplease or project level is there? You can use `git commit -n` or `git commit --no-verify` if you want to skip the commitplease  hook so that case is already possible with no extra work. It also accepts component names like `squash!` to help with rebasing against master. 

The only case I can see for not wanting commitplease is if you think contributors will be really confused when they try to commit and decide to ragequit rather than submit a PR. If that's happening we should improve the README and/or the error message. Otherwise we end up having to have a discussion with the contributor in the PR about the failed CI and asking them to rebase and fix it all, or do it for them.

dmethvin:
So let me be clear about what I think is strange for this situation. I think it's fine for this project to have different guidelines for what a commit message should look like. Since only the project maintainers land this stuff and we tend to rebase/squash commits, the maintainer has the ability to rewrite the message anyway and it hardly matters what the PR message says. 

The thing that seems odd in this case is to use the jQuery common format and only validate it when the PR is made. That means you'll be getting CI errors late in the process when it could have been found earlier.

dmethvin:
DAMMIT GITHUB

rxaviers:
Here's my understanding and conclusion of everything...

**Goal**

- The commits of this repo should follow http://contribute.jquery.org/commits-and-pull-requests/, specially before landing on master.
- Similarly, but unrelated to this PR, the commits of this repo should follow [style guides][], have License (CLA or CAA) approved, and do NOT introduce test regressions.

[style guides]: http://contribute.jquery.org/style-guide/

**Workflow for checking style guides, license, and test regressions**

For everything except for License, the workflow is to run `grunt`. It passes when all criteria is met. It fails when any criterion is unmet.

On local environment, contributors have the freedom to break tests, to not follow style guide, to commit all that, to do whatever they want while they develop. Contributors can run `grunt` locally to get assistance for the failures. So, they can fix them all before sending a PR.

On PRs, grunt is run by travis/CI and a nice and clean status is automatically set for every PR with whether or not all criteria has passed, which helps on triage and review.

**Commitplease git hook**

On local environment, the commitplease git hook prevents someone from performing a commit using invalid message.

On PRs, it's assumed that contributors have installed (and therefore used) commitplease. So, no invalid commit messages would be present. Although, that's not a guarantee. Reviewers cannot count with this. All statues could look green and incorrect commit messages still present. Therefore, it doesn't help on triage and review.

**Commitplease grunt task**

On local environment, contributors can run `grunt` locally to get assistance for all failures (now, including invalid commit messages as well). So, they can fix them all before sending a PR. it's the same workflow of everything else.

On PRs, it's again the same workflow of everything else. Grunt is run by travis/CI and a nice and clean status is automatically set for every PR with whether or not all criteria has passed (including invalid commit messages), which helps on triage and review.

**My conclusions**

The commitplease git hit hook enforcement may have helped other jQuery projects to drop the problem of invalid commit messages on its PRs. But, having a reliable green status on a PR to help the triage and review is a must.

Given (a) the linearity of using a grunt task approach compared to how we check other criteria and (b) the ability for checking invalid commit messages on both local and PRs/CI environments, we must give the grunt task approach a try first.

Given the invasive aspect of the git hook, because it prevents people from performing actions on their local environment, we should defer it for now. We don't want to prevent people from saving their files when jscs, or jshint, or a test fails. So, I don't think we should prevent people from committing their changes unless there's a strong evidence we must enforce that. Having said that, I still do think that:
- if contributors complain/request for a local tool/assistance to prevent them from committing invalid messages, we should explain them how to install the git hook themselves.
- if the previous complain becomes repetitive and hinders/prevents contributors from contributing to the project, we could consider enforcing the git hook on everyone.

PR #381, opened 2015-01-12
Build: Enable automated crossbrowser testing, including PRs

### LINKED ISSUE #235: Testing automation

jzaefferer:
Fixes #235

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,12 @@
 language: node_js
 node_js:
 - '0.10'
-before_install:
+install:
 - npm install -g grunt-cli
 - npm install -g bower
-install:
 - npm install
 - bower install
+- true && `base64 --decode <<< ZXhwb3J0IEJST1dTRVJTVEFDS19VU0VSTkFNRT1icm93c2Vyc3RhY2tnbG9iMQo=`
+- true && `base64 --decode <<< ZXhwb3J0IEJST1dTRVJTVEFDS19LRVk9TnBGNENONkFHVGp1RlhTTEV3Vk4K`
+script:
+- npm run-script ci
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
     "cldr-data": ">=25"
   },
   "devDependencies": {
+    "browserstack-runner": "0.3.0",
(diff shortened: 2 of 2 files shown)


jzaefferer:
This still isn't working properly. I can get runs to pass locally, but the last amend failed on Travis: https://travis-ci.org/jquery/globalize/builds/52004614

The error always involves "requirejs failed to load". It mostly happens for `unit.html`, but there's at least one `functional.html` failure as well. I wonder if increasing the requirejs `waitSeconds` option further would help. We've already changed it from the default 7 to 30.

jzaefferer:
I've rebased this after #431 landed, but now it fails due to the commitplease:currentBranch task. I have no idea what's up with that, will wait for Rafael to fix that in master and try again. @rxaviers would be great if you could also run browserstack-runner locally (make sure to install at least 0.3.0).

rxaviers:
I'm looking forward to having automated crossbrowser testing! :+1:

> I have no idea what's up with that

AFAICT, Travis clones the latest 50 commits only (for improved speed). Therefore, the commit range referenced by commitplease task is not available there.

If we can configure Travis to clone it up to the `--root`, this is a non-problem. By the way, cloning the whole repo in my Brazilian Internet connection takes 6s, which is very little compared to running `browserstack-runner` anyway.

> @rxaviers would be great if you could also run browserstack-runner locally (make sure to install at least 0.3.0).

Do you want me to test anything in particular in my local environment? Why do you want me to run it please?

jzaefferer:
> If we can configure Travis to clone it up to the --root, this is a non-problem. By the way, cloning the whole repo in my Brazilian Internet connection takes 6s, which is very little compared to running browserstack-runner anyway.

Sounds good.

> Do you want me to test anything in particular in my local environment? Why do you want me to run it please?

Just to avoid a "its working on my machine!!!".

rxaviers:
We'll need to update this according to #450 changes.

rxaviers:
Closing in favor of #450

PR #384, opened 2015-01-20
Use cldrjs 0.4.0 and Bundle Lookup Matcher

### LINKED ISSUE #357: CLDR data is reported as missing when locale is set with a region
When I try to use Globalize with a language-region combo (like en-US,) it always gives me the error: 
    `Error: E_MISSING_CLDR: Missing required CLDR content 'main/en/dates/calendars/gregorian/dateTimeFormats/short'.`

The basic code is as follows:

    Globalize.load(
    {
      "main": {
        "en-US": {
	    ... // content omitted
      }
    }
    });
    Globalize.locale('en-US');
    var formatter = Globalize.dateFormatter({datetime: 'short'});

At that point, globalize always throws the error.  However, if I provide it with CLDR data that is not region-dependent (like just plain 'en') and set my locale to the same, the code works.  In both cases, it is looking for cldr data that is just under the language code (ignoring the region for the cldr path.)  I have tested de-DE, and pt-BR and see the same results in those languages: it only looks for the language cldr data, ignoring the language-REGION data that was provided.

I saw an [issue](https://github.com/rxaviers/cldrjs/issues/17#issuecomment-48643041) on cldrjs that implied that the issue could be solved by using just en instead of en-US (which I am doing as a stopgap) but this same issue occurs for any region specific locale, and just using the language won't always be the right thing to do.

I'm not sure if this issue is better here or on cldrjs.  Let me know if I need to move it, or provide more information.

rxaviers:
Fixes #357

### DIFF
--- a/test/unit/core.js
+++ b/test/unit/core.js
@@ -74,6 +74,7 @@ QUnit.test( "should lookup bundle", function( assert ) {
 	// `en-US` bundle.
 	assert.equal( Globalize( "yy" ).cldr.attributes.bundle, "yy-YY" );
 	assert.equal( Globalize( "yy-YY" ).cldr.attributes.bundle, "yy-YY" );
+
 	assert.deepEqual( Globalize( "xx-XZ" ).cldr.attributes.bundle, null );
 });
 
--- a/test/unit/core.js
+++ b/test/unit/core.js
@@ -1,0 +1,80 @@
+define([
+	"cldr",
+	"src/core",
+	"json!cldr-data/supplemental/likelySubtags.json",
+
+	"cldr/event"
+], function( Cldr, Globalize, likelySubtags ) {
+
+Cldr.load( likelySubtags, {
+	supplemental: {
+		likelySubtags: {
+
+			// The below are fictitious locales for the purpose of this test.
+			// Note to self: don't use zz, Zzzz or ZZ. These are reserved subtags.
+			"xx": "xx-Xxxx-XX",
+			"yy": "yy-Xxxx-YY"
+		}
+	},
+	main: {
+		"de-CH": {},
+		en: {},
+		"en-GB": {},
+		"pa-Arab": {},
+		"sr-Cyrl": {},
+		"sr-Latn": {},
+		"uz-Arab": {},
+		"xx": {},
+		"yy-YY": {},
+		"zh": {},
+		"zh-Hant": {}
+	}
+});
+
+QUnit.module( "Globalize (constructor)" );
+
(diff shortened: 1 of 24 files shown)


jzaefferer:
That seems like a lot of changes for a simple version bump on a dependency. Do we really need to test the `attributes.bundle` property here? Isn't that already covered by tests in Cldrjs? (I don't know, I don't want to check since I'm not familiar with the Cldrjs codebase)

test/unit/core.js
```diff
@@ -0,0 +1,81 @@
...
+	assert.equal( Globalize( "sr-Latn" ).cldr.attributes.bundle, "sr-Latn" );
+	assert.equal( Globalize( "sr-Latn-RS" ).cldr.attributes.bundle, "sr-Latn" );
+	assert.equal( Globalize( "sr-RS" ).cldr.attributes.bundle, "sr-Cyrl" );
+	assert.equal( Globalize( "uz-AF" ).cldr.attributes.bundle, "uz-Arab" );
+	assert.equal( Globalize( "uz-Arab" ).cldr.attributes.bundle, "uz-Arab" );
+	assert.equal( Globalize( "zh" ).cldr.attributes.bundle, "zh" );
+	assert.equal( Globalize( "zh-CN" ).cldr.attributes.bundle, "zh" );
+	assert.equal( Globalize( "zh-Hans" ).cldr.attributes.bundle, "zh" );
+	assert.equal( Globalize( "zh-Hant" ).cldr.attributes.bundle, "zh-Hant" );
+	assert.equal( Globalize( "zh-TW" ).cldr.attributes.bundle, "zh-Hant" );
```
rxaviers: Note to self: Check `Globalize( locale ).main( path )` message instead of relying on `cldr.attributes.bundle` (cldrjs) internals.

rxaviers:
I agree we shouldn't test anything that is cldrjs internals, e.g., `attributes.bundle`. Although, I do think we must prevent regressions for any of our use cases, which includes making sure all those locales get the right data (independently of cldrjs tests).

rxaviers:
About the number of changes per commit... If you think we should break it down, just let me know how. Using cldrjs 0.4.0 means we're fixing Globalize by using its Bundle Lookup Matcher, not necessarily that it should be simple. Anyway, most of the changes are related to tests to prevent Globalize to regress #357, which is the reason for this change. So, I wouldn't split it.

test/unit/core.js
rxaviers: Updated

jzaefferer:
Looks good to me.

PR #386, opened 2015-01-28
Number rounding: Avoid inaccurate floating point arithmetics

### LINKED ISSUE #376: Incorrect decimal rounding by Globalize.format()
Hi,

I see that formatNumber function uses multiplication of the float by some power of 10 to round decimals. It works, except sometimes a decimal of 5 is rounded down instead of up. For example: 
Globalize.format(0.00015, "n4") // Returns 0.0001, should return 0.0002.
Globalize.format(1.00015, "n4") // Return 1.0002 correctly.

A better solution would be to use numbers represented in exponential notation:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/round

Cheers!

rxaviers:
JavaScript has a set of native math methods to perform number rounding: Math.ceil, Math.floor and Math.round. All of them return the number rounded to its nearest (ceil, floor or round) integer.

Rounding a decimal number on JavaScript is a three step process: shifting the decimal to a desired integer, rounding it and shifting the rounded number back to its original decimal order.

Currently, arithmetical operations `*` and `/` are used for shifting a number and for shifting it back. It happens these are subject to floating point inaccuracies. For example:
- `0.00015 * 10000 = 1.4999999999999998`;
- `1234 * 0.0001 = 0.12340000000000001`

Inaccurate floating point arithmetics lead to inaccurate roundings. In order to mitigate them, this change implements a custom way of performing both shifting and shifting back operations. Instead of multiplying the numbers by powers of 10, their floating point components are operated directly (summing or subtracting their exponents). Based on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/round and suggested by @stonys (Andrius Stonys).

Fixes #376

### DIFF
--- a/src/util/number/round.js
+++ b/src/util/number/round.js
@@ -7,11 +7,12 @@ define([
  *
  * @method [String] with either "round", "ceil", "floor", or "truncate".
  *
- * Return function( value, increment ):
+ * Return function( value, incrementOrExp ):
  *
  *   @value [Number] eg. 123.45.
  *
- *   @increment [Number] optional, eg. 0.1.
+ *   @incrementOrExp [Number] optional, eg. 0.1; or
+ *     [Object] Either { increment: <value> } or { exponent: <value> }
  *
  *   Return the rounded number, eg:
  *   - round( "round" )( 123.45 ): 123;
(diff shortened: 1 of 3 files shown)


src/util/number/round.js
```diff
@@ -20,13 +21,67 @@ define([
...
 	method = method || "round";
 	method = method === "truncate" ? numberTruncate : Math[ method ];
-	return function( value, increment ) {
-		increment = increment || 1;
-		return method( value / increment ) * increment;
+
+	return function( value, incrementOrExp ) {
+		var exp, increment;
+
+		value = +value;
```
jzaefferer: This conversion is new. Considering the limitations of `isNaN`, should you first check if `value` is `typeof number`, and only then use `isNaN`?

test/unit/number/format.js
```diff
@@ -151,6 +151,10 @@ QUnit.test( "should allow rounding", function( assert ) {
 	assert.equal( format( pi, properties( "0.20", en ) ), "3.20" );
 	assert.equal( format( pi, properties( "0.5", en ) ), "3.0" );
 	assert.equal( format( pi, properties( "0.1", en ) ), "3.1" );
+
+	// Handle inaccurate floating point arithmetics like 0.00015 * 10000 = 1.49999999999999.
+	// See #376.
+	assert.equal( format( 0.00015, properties( "0.0001", en ) ), "0.0002" );
```
jzaefferer: This looks like very little additional coverage relative to the complexity of the changes above. Is there no effect on other methods besides rounding? They all seem to share the same code path.

jzaefferer:
Except for a few details that may just be a lack of understanding on my end, this looks good. I haven't fully dug into the algorithm, but since its more or less a copy from the MDN article, that should be fine.

test/unit/number/format.js
rxaviers: This change fixes inaccurate floating point arithmetics related to rounding decimals only. its usage is not a special option, it's always used when formatting a number/currency with decimal digits. Any rounding that is performed by number/currency formatting is performed by the same rounding method that I'm testing above. Did I answer you? Let me ask + answer some other related questions myself...

*Can other parts of the globalize code eventually run into the same rounding problem reported here and we are not testing it?*

No. Any rounding is performed by the method I'm testing above. This is true at least for any number code. On date parsing, the native Math.round is used directly to parse milliseconds and the native Math.floor is used directly to parse centuries, which I am not sure the same precaution should be taken. Any test in mind for those?

*Can this test be placed somewhere else?*

Yes. I've placed this test under "Number Format ld allow rounding". But, this test could be placed either under maximum fraction digits, or under minimum fraction, or under significant digits formatting, etc. Formatting a decimal in any of those forms would eventually hit the same problem. But, testing it in one place is enough to catch it.

*Other globalize code indirectly uses the rounding code that has been changed. Is everything else working just fine like before?*

Yes, it should. At least, it didn't regress in any test. All tests are still passing, including `1234 * 0.0001 = 0.12340000000000001` case tested by previous existing "Number Format should format significant digits".

Please, just let me know if you still have any question.

src/util/number/round.js
rxaviers: > This conversion is new.

Please, could you explain? The conversion is present on Mozilla's code and it's also already been used throughout globalize code.

> Considering the limitations of isNaN, should you first check if value is typeof number, and only then use isNaN?

According to our talk, using isNaN as-is is fine in this context.

jzaefferer:
Alright, thanks for the clarifications. Let's merge it then!

Issue #205, opened 2014-01-31
Support relative time


rxaviers:
Support relative times:
- in the past , eg "5 days ago";
- in the future, eg "in 5 days";

#### TODO
- [ ] Implementation;
- [ ] Unit tests;
- [ ] Functional tests;
- [ ] Docs;

#### API

```javascript
Globalize.formatDuration( value[, options] )
```

Examples:

```javascript
.formatDuration( inTwoDays ); // in 2 days
.formatDuration( twoDaysAgo ); // 2 days ago
.formatDuration( twoDaysAgo, { unit: "hour" }); // 48 hours ago
.formatDuration({ from: christmas2001, to: newyear2002 }, { unit: "day" }); // in 7 days
```

Value: Date instance, or object with `from` and `to` keys and Date instance values.

Options:
- unit ("day"|"week"|"month"|...), eg. "in 30 days", "in 4 weeks", or "in 1 month", ... Defaults to automatic detection, therefore implementing a simple approximation algorithm as Jörn suggested. But, delegating any more complex things to a specific date manipulation library such as moment.js.

Returns:
- "1 second ago" .. "59 seconds ago"; (1 .. 59 s)
- "1 minute ago" .. "59 minutes ago"; (60 .. 60*60 s)
- "1 hour ago" .. "59 hours ago";
- "1 day ago" .. "59 days ago";
- "1 week ago" .. "59 weeks ago";
- "1 month ago" .. "59 months ago";
- "1 year ago" .. "59 years ago";

(analogous for future)

#### CLDR

Relative time fields can be found on <lang>/[dateFields.json](http://www.unicode.org/repos/cldr-aux/json/24/en/dateFields.json).
- http://www.unicode.org/reports/tr35/tr35-dates.html#Calendar_Fields
- http://www.unicode.org/reports/tr35/tr35-dates.html#Date_Field_Symbol_Table

#### Other

References:
- [twitter-cldr]
- [Rails](http://api.rubyonrails.org/classes/ActionView/Helpers/DateHelper.html#method-i-distance_of_time_in_words);
- Wikipedia [[1]](https://github.com/wikimedia/mediawiki-core/blob/master/includes/MWTimestamp.php#L179) [[2]](https://github.com/wikimedia/mediawiki-core/blob/master/languages/Language.php#L2269);

rxaviers:
[...]

PR #389, opened 2015-02-11
Add relative time formatting

mbirtwell:
Hi,

I wrote a couple of functions for formatting some date related things based on the cldr data. 
getDay just looks up the names of days in the cldr data. (and supports the different formats that cldr supports)
formatRelativeCount formats an amount of time in a relative way. e.g. if you ask for 3 days in the past it will return 3 days ago. It tries to use words like today and tomorrow if available.

### DIFF
--- a/src/build/intro-date.js
+++ b/src/build/intro-date.js
@@ -48,4 +48,5 @@ var createError = Globalize._createError,
 	validateDefaultLocale = Globalize._validateDefaultLocale,
 	validateParameterPresence = Globalize._validateParameterPresence,
 	validateParameterType = Globalize._validateParameterType,
-	validateParameterTypeString = Globalize._validateParameterTypeString;
+	validateParameterTypeString = Globalize._validateParameterTypeString,
+    validateParameterTypeNumber = Globalize._validateParameterTypeNumber;
--- a/src/build/intro-date.js
+++ b/src/build/intro-date.js
@@ -49,4 +49,4 @@ var createError = Globalize._createError,
 	validateParameterPresence = Globalize._validateParameterPresence,
 	validateParameterType = Globalize._validateParameterType,
 	validateParameterTypeString = Globalize._validateParameterTypeString,
-    validateParameterTypeNumber = Globalize._validateParameterTypeNumber;
+	validateParameterTypeNumber = Globalize._validateParameterTypeNumber;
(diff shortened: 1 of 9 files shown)


rxaviers:
@mbirtwell, thanks for your contribution.

I will review your code soon. But, quick comments.
 1. Could you read&sign our [CLA](http://contribute.jquery.org/CLA/)?
 1. Could you rewrite your `formatRelativeCount` according to https://github.com/jquery/globalize/issues/205 (see also https://github.com/jquery/globalize/tree/fix-205-format-duration)?

mbirtwell:
I've signed the CLA
I'll have a look at converting the work that I've done for formatRelativeCount in to formatDuration. It looks like #205 potentially covers a lot more functionality than I have attempted to implement. So I'll start by trying to re-organising what I've done so far as clean subset of the functionality and let you know how I get on.

mbirtwell:
Ok I think I've got my head around most of the conversation in #205 what I propose doing is closing this pull request and creating two new ones.

One for getDay because I think that's useful as it is. Although arguably could live in cldrjs. 

One for a relativeDuration function (the name is flexible), which will have the same API as my formatRelativeCount function which I think is very similar to the value + unit low level api that @ichernev  is proposing. A full implementation could then be built on top of this. Unfortunately I want this api not the one that takes dates, so I'm not going to implement that.

Let me know what you think?

rxaviers:
> getDay just looks up the names of days in the cldr data

This method is very similar to `.formatDate( <date>, "E" )`. It's also very specific (you would end up having another method for Month names, quarter names, era names and so on). Also, it can be replaced by one line using cldrjs `cldr.main([ "dates/calendars/gregorian/days/stand-alone", format, day ]);`. So, I'd recommend one to either use `.formatDate()` or cldrjs directly on that case.

rxaviers:
About the relative time format function, on #205 we came up with an API that has Dates as parameters (which is summarized in the description of #205). You are suggesting it should take integer values. I think we can extend our original API (that takes Dates) to accept either a Date or an integer value.

I suggest you to use our existing branch as a baseline and continue implementing from there. You could either close this PR and create a new one, or update this PR. I'm fine with either options.

If you are interested, I can rebase our existing branch to the current master (in order to get that up-to-date first). How does that sound to you?

mbirtwell:
I had a go at merging master in your fix-205 branch earlier and got that to
the point of passing grunt. So I think I can continue from there.

I think there really is two functions here one that takes dates and on that
takes integer and units. I'd imagine the one that takes dates would be
implemented using the other but it seems to me trying to combine them makes
the interface a bit confusing.

rxaviers:
Superseded by #395

Issue #272, opened 2014-06-03
Date (skeletons): handle date and time in the same skeleton


rxaviers:
Currently, it's possible to:

```
> Globalize.formatDate(new Date(), "MMMd")
'Jun 3'
> Globalize.formatDate(new Date(), "hms")
'1:40:31 PM'
```

Although, `.formatDate( date, "MMMdhm" )` isn't. But, it should.
http://www.unicode.org/reports/tr35/tr35-dates.html#availableFormats_appendItems

> If a client-requested set of fields includes both date and time fields, and if the availableFormats data does not include a dateFormatItem whose skeleton matches the same set of fields, then the request should be handled as follows:

> 1. Divide the request into a date fields part and a time fields part
> 2. For each part, find the matching dateFormatItem
> 3. Combine the patterns for the two dateFormatItems using the appropriate dateTimeFormat pattern, determined as follows from the requested date fields:
>  - If the requested date fields include wide month (MMMM, LLLL) and weekday name of any length (e.g. E, EEEE, c, cccc), use <dateTimeFormatLength type="full">
>  - Otherwise, if the requested date fields include wide month, use <dateTimeFormatLength type="long">
>  - Otherwise, if the requested date fields include abbreviated month (MMM, LLL), use <dateTimeFormatLength type="medium">
>  - Otherwise use <dateTimeFormatLength type="short">

See also: https://github.com/ibm-js/ecma402/issues/76

ManrajGrover:
@rxaviers sir,

I see [ibm-js/ecma402/issues/76](https://github.com/ibm-js/ecma402/issues/76) have already implemented `long` , `medium` , `short` which is already available here under preset `datetime` . If there is more customization like above examples given by you, what we can do is parse the skeleton and see date and time component and call formatDate for both separately and return accordingly. But yes there may be several cases we have to take care of. For example, `"GyMMM"` and similar skeletons: In this we would need to put `AD` after the time. Also it may be possible user asks for time first before date. So should take care of all flexibilities.

I would like to take this up if possible.

rxaviers:
> is already available here under preset datetime

Yeap, this issue isn't about the presets. But, skeletons.

> For example, "GyMMM" and similar skeletons: In this we would need to put AD after the time. Also it may be possible user asks for time first before date. So should take care of all flexibilities.

[...]

Issue #271, opened 2014-06-03
Date (skeletons): dynamically augment availableFormats


rxaviers:
WIP PR: #703 (replaced  ~~#604~~  and ~~#462~~)

---

As specification says:

> It is not necessary to supply dateFormatItems with skeletons for every field length; in many cases fields in the skeleton and pattern can be expanded in parallel to handle a request. Consider the following dateFormatItem:
> 
> `<dateFormatItem id="yMMMd">d MMM y</dateFormatItem>`
> 
> If the request is for yMMMMd, the above item can be automatically expanded to produce the pattern "d MMMM y" in response to the request. Of course, if the desired behavior is that a request for yMMMMd should produce something other than "d MMMM y", a separate dateFormatItem must be added, for example:
> 
> `<dateFormatItem id="yMMMMd">d 'de' MMMM 'de' y</dateFormatItem>`

http://www.unicode.org/reports/tr35/tr35-dates.html#availableFormats_appendItems

We need to support that on `.formatDate()`.

AndreR:
Any news on this topic? I wanted to use the skeleton "MMMM" with Globalize set to German, but I ended up with that exact same error message _E_UNSUPPORTED: Unsupported {"feature": "year pattern `u`"}_. If I set it to English, it works, though.

Any chance that this will be fixed anytime soon?

rxaviers:
Hi @AndreR, #604 is stalled. Would you like to take it over? I'd be happy if you could address the comments in that PR. You could create a new pull request on top of that one with additional commits. Just let me know if you have any questions. Thanks

AndreR:
Hi @rxaviers, I would love to help you out here, but my Javascript skills are too basic. I'm sorry!

PR #426, opened 2015-03-18
Fixed issue 272 Handle DateTime in same skeleton.

ManrajGrover:
@rxaviers sir,

Please review the code for issue #272 . I have tested it by making changes in `date.js` but not in `expand-pattern.js` . Please let me know the changes.

### DIFF
--- a/src/date.js
+++ b/src/date.js
@@ -27,8 +27,8 @@ define([
 function validateRequiredCldr( path, value ) {
 	validateCldr( path, value, {
 		skip: [
-			/dates\/calendars\/gregorian\/days\/.*\/short/,
 			/dates\/calendars\/gregorian\/dateTimeFormats\/availableFormats/,
+			/dates\/calendars\/gregorian\/days\/.*\/short/,
 			/supplemental\/timeData\/(?!001)/,
 			/supplemental\/weekData\/(?!001)/
 		]
--- a/src/date.js
+++ b/src/date.js
@@ -28,9 +28,9 @@ function validateRequiredCldr( path, value ) {
 	validateCldr( path, value, {
 		skip: [
 			/dates\/calendars\/gregorian\/days\/.*\/short/,
+			/dates\/calendars\/gregorian\/dateTimeFormats\/availableFormats/,
 			/supplemental\/timeData\/(?!001)/,
-			/supplemental\/weekData\/(?!001)/,
-			/dates\/calendars\/gregorian\/dateTimeFormats\/availableFormats/
+			/supplemental\/weekData\/(?!001)/
 		]
 	});
 }
--- a/src/date.js
+++ b/src/date.js
@@ -29,7 +29,8 @@ function validateRequiredCldr( path, value ) {
 		skip: [
 			/dates\/calendars\/gregorian\/days\/.*\/short/,
 			/supplemental\/timeData\/(?!001)/,
-			/supplemental\/weekData\/(?!001)/
+			/supplemental\/weekData\/(?!001)/,
+			/dates\/calendars\/gregorian\/dateTimeFormats\/availableFormats/
 		]
 	});
 }
--- a/src/date/expand-pattern.js
+++ b/src/date/expand-pattern.js
@@ -1,24 +1,3 @@
-function getDateTime( type, cldr ) {
-	var result;
-	result = cldr.main([
-		"dates/calendars/gregorian/dateTimeFormats",
-		type
-	]);
-	if ( result ) {
-		result = formatMessage( result, [
-			cldr.main([
-				"dates/calendars/gregorian/timeFormats",
-				type
-			]),
-			cldr.main([
-				"dates/calendars/gregorian/dateFormats",
-				type
-			])
-		]);
-	}
-	return result;
-}
-
 define([
 	"../common/format-message",
 	"../common/create-error/invalid-parameter-value"
@@ -46,6 +25,27 @@ define([
  * - { pattern: "dd/mm" } returns "dd/mm";
  */
 
+function getDateTime( type, dateSkeleton, timeSkeleton, cldr ) {
+	var result;
+	result = cldr.main([
+		"dates/calendars/gregorian/dateTimeFormats",
+		type
+	]);
+	if ( result ) {
+		result = formatMessage( result, [
+			cldr.main([
+				"dates/calendars/gregorian/dateTimeFormats/availableFormats",
+				timeSkeleton
+			]),
+			cldr.main([
+				"dates/calendars/gregorian/dateTimeFormats/availableFormats",
+				dateSkeleton
+			])
+		]);
+	}
+	return result;
+}
+
 return function( pattern, cldr ) {
 	var result;
 
@@ -57,14 +57,17 @@ return function( pattern, cldr ) {
 		case "skeleton" in pattern:
 			var flagDate = false, flagTime = false;
 			var skeleton = pattern.skeleton;
-			for ( var i = 0 ; i < skeleton.length ; i++ ) {
-				var ch = skeleton[ i ];
-				if ( ch == 'h' || ch == 'H' || ch == 'm' || ch == 's' ) { //Check if skeleton contains Time component
-					flagTime = true;
-				}
-				if ( ch == 'y' || ch == 'M' || ch == 'd' ) { // Check if skeleton contains Date component
-					flagDate = true;
-				}
+			if ( /[hHms]/.test( skeleton ) ) {
+				flagTime = true;
+			}
+			if ( /[GyQMd]/.test( skeleton ) ) {
+				flagDate = true;
+			}
+			if ( /Ed/g.test( skeleton ) ) {
+				flagDate = true;
+			}
+			else if ( /Eh/g.test( skeleton ) || /EH/g.test( skeleton ) || /E/g.test( skeleton ) ) {
+				flagTime = true;
 			}
 			if ( ( flagDate != flagTime ) ) {
 				result = cldr.main([
@@ -73,24 +76,55 @@ return function( pattern, cldr ) {
 				]);
 			}
 			else if ( flagDate && flagTime ) {
-				if ( skeleton.indexOf( "MMMM" ) > -1 && skeleton.indexOf( "E" ) > -1 ) {
-					result = getDateTime( "full", cldr );
+				var dateSkeleton = "", timeSkeleton = "";
+				for ( var i = 0 ; i < skeleton.length ; i++ ) {
+					// Get timeSkeleton
+					if ( /[hHms]/.test( skeleton[i] ) ) {
+						timeSkeleton += skeleton[i];
+					}
+					// Get dateSkeleton
+					else if ( /[GyQMd]/.test( skeleton[i] ) ) {
+						dateSkeleton += skeleton[i];
+					}
+					// Check for "E"
+					else if ( skeleton[ i ] == "E" ) {
+						if ( /Ed/g.test( skeleton.slice( i ) ) ) {
+							dateSkeleton += "Ed";
+							i++;
+						}
+						else if ( /Eh/g.test( skeleton.slice( i ) ) ) {
+							timeSkeleton += "Eh";
+							i++;
+						}
+						else if ( /EH/g.test( skeleton.slice( i ) ) ) {
+							timeSkeleton += "EH";
+							i++;
+						}
+						else {
+							timeSkeleton += "E";
+						}
+					}
 				}
-				else if ( skeleton.indexOf( "MMMM" ) > -1 ) {
-					result = getDateTime( "long", cldr );
+				if ( /MMMM/g.test( dateSkeleton ) && /[E]/g.test( dateSkeleton ) ) {
+					result = getDateTime( "full", dateSkeleton, timeSkeleton, cldr);
 				}
-				else if ( skeleton.indexOf( "MMM" ) > -1 ) {
-					result = getDateTime( "medium", cldr );
+				else if ( /MMMM/g.test( dateSkeleton ) ) {
+					result = getDateTime( "long", dateSkeleton, timeSkeleton, cldr);
+				}
+				else if ( /MMM/g.test( dateSkeleton ) ) {
+					result = getDateTime( "medium", dateSkeleton, timeSkeleton, cldr);
+					if ( result.indexOf("G") > -1 ) {
+						result = result.replace('G','');
+						result += " G";
+					}
 				}
 				else {
-					result = getDateTime( "short", cldr );
+					result = getDateTime( "short", dateSkeleton, timeSkeleton, cldr);
+					if ( result.indexOf("G") > -1 ) {
+						result = result.replace('G','');
+						result += " G";
+					}
 				}
-			}
-			else {
-				result = cldr.main([
-					"dates/calendars/gregorian/dateTimeFormats/availableFormats",
-					skeleton
-				]);
 			}
 			break;
 
--- a/src/date/expand-pattern.js
+++ b/src/date/expand-pattern.js
@@ -26,9 +26,9 @@ define([
  */
 
 return function( pattern, cldr ) {
-	var result, getDateTime, skeleton, dateSkeleton = "", timeSkeleton = "";
+	var result, getDateTime, skeleton, dateSkeleton = "", timeSkeleton = "", type;
 
-	getDateTime = function( type, dateSkeleton, timeSkeleton, cldr ) {
+	getDateTime = function( type, dateSkeleton, timeSkeleton, datePath, timePath, cldr ) {
 		var result;
 		result = cldr.main([
 			"dates/calendars/gregorian/dateTimeFormats",
@@ -37,11 +37,11 @@ return function( pattern, cldr ) {
 		if ( result ) {
 			result = formatMessage( result, [
 				cldr.main([
-					"dates/calendars/gregorian/dateTimeFormats/availableFormats",
+					timePath,
 					timeSkeleton
 				]),
 				cldr.main([
-					"dates/calendars/gregorian/dateTimeFormats/availableFormats",
+					datePath,
 					dateSkeleton
 				])
 			]);
@@ -64,14 +64,17 @@ return function( pattern, cldr ) {
 				timeSkeleton = skeleton.split( /[^hHKkmsSAzZOvVXx]/ ).slice( -1 )[ 0 ];
 				dateSkeleton = skeleton.split( /[^GyYuUrQqMLlwWdDFgEec]/ )[ 0 ];
 				if (  /(MMMM|LLLL).*[Ec]/.test( dateSkeleton ) ) {
-					result = getDateTime( "full", dateSkeleton, timeSkeleton, cldr );
+					type = "full";
 				} else if ( /MMMM/g.test( dateSkeleton ) ) {
-					result = getDateTime( "long", dateSkeleton, timeSkeleton, cldr );
+					type = "long";
 				} else if ( /MMM/g.test( dateSkeleton ) || /LLL/g.test( dateSkeleton ) ) {
-					result = getDateTime( "medium", dateSkeleton, timeSkeleton, cldr );
+					type = "medium";
 				} else {
-					result = getDateTime( "short", dateSkeleton, timeSkeleton, cldr );
+					type = "short";
 				}
+				result = getDateTime( type, dateSkeleton, timeSkeleton,
+						"dates/calendars/gregorian/dateTimeFormats/availableFormats",
+						"dates/calendars/gregorian/dateTimeFormats/availableFormats", cldr );
 			}
 			break;
 
@@ -85,22 +88,9 @@ return function( pattern, cldr ) {
 			break;
 
 		case "datetime" in pattern:
-			result = cldr.main([
-				"dates/calendars/gregorian/dateTimeFormats",
-				pattern.datetime
-			]);
-			if ( result ) {
-				result = formatMessage( result, [
-					cldr.main([
-						"dates/calendars/gregorian/timeFormats",
-						pattern.datetime
-					]),
-					cldr.main([
-						"dates/calendars/gregorian/dateFormats",
-						pattern.datetime
-					])
-				]);
-			}
+			result = getDateTime( pattern.datetime, pattern.datetime, pattern.datetime,
+					"dates/calendars/gregorian/dateFormats",
+					"dates/calendars/gregorian/timeFormats", cldr );
 			break;
 
 		case "pattern" in pattern:
--- a/src/date/expand-pattern.js
+++ b/src/date/expand-pattern.js
@@ -66,7 +66,8 @@ return function( pattern, cldr ) {
 			if ( /Ed/g.test( skeleton ) ) {
 				flagDate = true;
 			}
-			else if ( /Eh/g.test( skeleton ) || /EH/g.test( skeleton ) || /E/g.test( skeleton ) ) {
+			else if ( /Eh/g.test( skeleton ) || /EH/g.test( skeleton ) || 
+					/E/g.test( skeleton ) ) {
 				flagTime = true;
 			}
 			if ( ( flagDate != flagTime ) ) {
--- a/src/date/expand-pattern.js
+++ b/src/date/expand-pattern.js
@@ -66,7 +66,7 @@ return function( pattern, cldr ) {
 			if ( /Ed/g.test( skeleton ) ) {
 				flagDate = true;
 			}
-			else if ( /Eh/g.test( skeleton ) || /EH/g.test( skeleton ) || 
+			else if ( /Eh/g.test( skeleton ) || /EH/g.test( skeleton ) ||
 					/E/g.test( skeleton ) ) {
 				flagTime = true;
 			}
--- a/src/date/expand-pattern.js
+++ b/src/date/expand-pattern.js
@@ -25,30 +25,30 @@ define([
  * - { pattern: "dd/mm" } returns "dd/mm";
  */
 
-function getDateTime( type, dateSkeleton, timeSkeleton, cldr ) {
-	var result;
-	result = cldr.main([
-		"dates/calendars/gregorian/dateTimeFormats",
-		type
-	]);
-	if ( result ) {
-		result = formatMessage( result, [
-			cldr.main([
-				"dates/calendars/gregorian/dateTimeFormats/availableFormats",
-				timeSkeleton
-			]),
-			cldr.main([
-				"dates/calendars/gregorian/dateTimeFormats/availableFormats",
-				dateSkeleton
-			])
-		]);
-	}
-	return result;
-}
-
 return function( pattern, cldr ) {
-	var result, skeleton, dateSkeleton = "", timeSkeleton = "";
-
+	var result, getDateTime, skeleton, dateSkeleton = "", timeSkeleton = "";
+	
+	getDateTime = function( type, dateSkeleton, timeSkeleton, cldr ) {
+		var result;
+		result = cldr.main([
+			"dates/calendars/gregorian/dateTimeFormats",
+			type
+		]);
+		if ( result ) {
+			result = formatMessage( result, [
+				cldr.main([
+					"dates/calendars/gregorian/dateTimeFormats/availableFormats",
+					timeSkeleton
+				]),
+				cldr.main([
+					"dates/calendars/gregorian/dateTimeFormats/availableFormats",
+					dateSkeleton
+				])
+			]);
+		}
+		return result;
+	};
+	
 	if ( typeof pattern === "string" ) {
 		pattern = { skeleton: pattern };
 	}
--- a/src/date/expand-pattern.js
+++ b/src/date/expand-pattern.js
@@ -68,8 +68,8 @@ return function( pattern, cldr ) {
 						dateSkeleton += skeleton[i];
 					}
 				}
-				if ( ( /MMMM/g.test( dateSkeleton ) || /LLLL/g.test( dateSkeleton ) ) 
-					&& /[Ec]/.test( dateSkeleton ) ) {
+				if ( ( /MMMM/g.test( dateSkeleton ) || /LLLL/g.test( dateSkeleton ) ) &&
+						/[Ec]/.test( dateSkeleton ) ) {
 					result = getDateTime( "full", dateSkeleton, timeSkeleton, cldr);
 				} else if ( /MMMM/g.test( dateSkeleton ) ) {
 					result = getDateTime( "long", dateSkeleton, timeSkeleton, cldr);
--- a/src/date/expand-pattern.js
+++ b/src/date/expand-pattern.js
@@ -27,7 +27,7 @@ define([
 
 return function( pattern, cldr ) {
 	var result, getDateTime, skeleton, dateSkeleton = "", timeSkeleton = "";
-	
+
 	getDateTime = function( type, dateSkeleton, timeSkeleton, cldr ) {
 		var result;
 		result = cldr.main([
@@ -48,7 +48,7 @@ return function( pattern, cldr ) {
 		}
 		return result;
 	};
-	
+
 	if ( typeof pattern === "string" ) {
 		pattern = { skeleton: pattern };
 	}
--- a/src/date/expand-pattern.js
+++ b/src/date/expand-pattern.js
@@ -60,23 +60,17 @@ return function( pattern, cldr ) {
 				"dates/calendars/gregorian/dateTimeFormats/availableFormats",
 				skeleton
 			]);
-			if ( result === undefined ) {
-				for ( i = 0 ; i < skeleton.length ; i++ ) {
-					if ( /[hHms]/.test( skeleton[i] ) ) {
-						timeSkeleton += skeleton[i];
-					} else if ( /[GyQMEd]/.test( skeleton[i] ) ) {
-						dateSkeleton += skeleton[i];
-					}
-				}
-				if ( ( /MMMM/g.test( dateSkeleton ) || /LLLL/g.test( dateSkeleton ) ) &&
-						/[Ec]/.test( dateSkeleton ) ) {
-					result = getDateTime( "full", dateSkeleton, timeSkeleton, cldr);
+			if ( !result ) {
+				timeSkeleton = skeleton.split( /[^hHKkmsSAzZOvVXx]/ ).slice( -1 )[ 0 ];
+				dateSkeleton = skeleton.split( /[^GyYuUrQqMLlwWdDFgEec]/ )[ 0 ];
+				if (  /(MMMM|LLLL).*[Ec]/.test( dateSkeleton ) ) {
+					result = getDateTime( "full", dateSkeleton, timeSkeleton, cldr );
 				} else if ( /MMMM/g.test( dateSkeleton ) ) {
-					result = getDateTime( "long", dateSkeleton, timeSkeleton, cldr);
+					result = getDateTime( "long", dateSkeleton, timeSkeleton, cldr );
 				} else if ( /MMM/g.test( dateSkeleton ) || /LLL/g.test( dateSkeleton ) ) {
-					result = getDateTime( "medium", dateSkeleton, timeSkeleton, cldr);
+					result = getDateTime( "medium", dateSkeleton, timeSkeleton, cldr );
 				} else {
-					result = getDateTime( "short", dateSkeleton, timeSkeleton, cldr);
+					result = getDateTime( "short", dateSkeleton, timeSkeleton, cldr );
 				}
 			}
 			break;
--- a/src/date/expand-pattern.js
+++ b/src/date/expand-pattern.js
@@ -88,7 +88,7 @@ return function( pattern, cldr ) {
 						dateSkeleton += skeleton[ i ];
 					}
 					// Check for "E"
-					else if ( skeleton[ i ] == "E" ) {
+					else if ( skeleton[ i ] === "E" ) {
 						if ( /Ed/g.test( skeleton.slice( i ) ) ) {
 							dateSkeleton += "Ed";
 							i++;
--- a/src/date/expand-pattern.js
+++ b/src/date/expand-pattern.js
@@ -47,7 +47,7 @@ function getDateTime( type, dateSkeleton, timeSkeleton, cldr ) {
 }
 
 return function( pattern, cldr ) {
-	var result, flagDate, flagTime, skeleton, dateSkeleton, timeSkeleton;
+	var result, flagDate, flagTime, skeleton, dateSkeleton, timeSkeleton, i;
 
 	if ( typeof pattern === "string" ) {
 		pattern = { skeleton: pattern };
@@ -76,7 +76,7 @@ return function( pattern, cldr ) {
 				]);
 			} else if ( flagDate && flagTime ) {
 				dateSkeleton = "", timeSkeleton = "";
-				for ( var i = 0 ; i < skeleton.length ; i++ ) {
+				for ( i = 0 ; i < skeleton.length ; i++ ) {
 					if ( /[hHms]/.test( skeleton[ i ] ) ) {
 						timeSkeleton += skeleton[ i ];
 					} else if ( /[GyQMd]/.test( skeleton[ i ] ) ) {
@@ -109,7 +109,7 @@ return function( pattern, cldr ) {
 				} else {
 					result = getDateTime( "short", dateSkeleton, timeSkeleton, cldr );
 					if ( result.indexOf( "G" ) > -1 ) {
-						result = result.replace( "G" , "" );
+						result = result.replace( "G", "" );
 						result += " G";
 					}
 				}
--- a/src/date/expand-pattern.js
+++ b/src/date/expand-pattern.js
@@ -105,7 +105,7 @@ return function( pattern, cldr ) {
 						}
 					}
 				}
-				if ( /MMMM/g.test( dateSkeleton ) && /[E]/g.test( dateSkeleton ) ) {
+				if ( /MMMM/g.test( dateSkeleton ) && /[E]/.test( dateSkeleton ) ) {
 					result = getDateTime( "full", dateSkeleton, timeSkeleton, cldr);
 				}
 				else if ( /MMMM/g.test( dateSkeleton ) ) {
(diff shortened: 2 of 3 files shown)


src/date/expand-pattern.js
```diff
@@ -33,10 +55,43 @@ return function( pattern, cldr ) {
 
 	switch ( true ) {
 		case "skeleton" in pattern:
-			result = cldr.main([
-				"dates/calendars/gregorian/dateTimeFormats/availableFormats",
-				pattern.skeleton
-			]);
```
rxaviers: > and if the availableFormats data does not include a dateFormatItem whose skeleton matches the same set of fields

First, check for availableFormats. Therefore, keep the above. If `result === undefined`, keep going...

src/date/expand-pattern.js
```diff
@@ -33,10 +55,43 @@ return function( pattern, cldr ) {
...
+			var flagDate = false, flagTime = false;
+			var skeleton = pattern.skeleton;
+			for ( var i = 0 ; i < skeleton.length ; i++ ) {
+				var ch = skeleton[ i ];
+				if ( ch == 'h' || ch == 'H' || ch == 'm' || ch == 's' ) { //Check if skeleton contains Time component
+					flagTime = true;
+				}
+				if ( ch == 'y' || ch == 'M' || ch == 'd' ) { // Check if skeleton contains Date component
+					flagDate = true;
+				}
```
rxaviers: > If a client-requested set of fields includes both date and time fields

1. What are all the available date fields?
1. What are all the available time fields?
1. A could be written as B:

A:
```js
ch == 'h' || ch == 'H' || ch == 'm' || ch == 's'
```

B:
```js
/[hHms]/.test( ch )
```

src/date/expand-pattern.js
```diff
@@ -33,10 +55,43 @@ return function( pattern, cldr ) {
...
+			}
+			if ( ( flagDate != flagTime ) ) {
+				result = cldr.main([
+					"dates/calendars/gregorian/dateTimeFormats/availableFormats",
+					skeleton
+				]);
+			}
+			else if ( flagDate && flagTime ) {
+				if ( skeleton.indexOf( "MMMM" ) > -1 && skeleton.indexOf( "E" ) > -1 ) {
```
rxaviers: > - If the requested date fields include wide month (MMMM, LLLL) and weekday name of any length (e.g. E, EEEE, c, cccc), use <dateTimeFormatLength type="full">

What about `L` and `c`?

src/date/expand-pattern.js
```diff
@@ -33,10 +55,43 @@ return function( pattern, cldr ) {
...
+					result = getDateTime( "full", cldr );
+				}
+				else if ( skeleton.indexOf( "MMMM" ) > -1 ) {
```
rxaviers: >  - Otherwise, if the requested date fields include wide month, use <dateTimeFormatLength type="long">

What about L?

src/date/expand-pattern.js
```diff
@@ -33,10 +55,43 @@ return function( pattern, cldr ) {
...
+					result = getDateTime( "long", cldr );
+				}
+				else if ( skeleton.indexOf( "MMM" ) > -1 ) {
```
rxaviers: >  - Otherwise, if the requested date fields include abbreviated month (MMM, LLL), use `<dateTimeFormatLength type="medium">`

What about L?

src/date/expand-pattern.js
```diff
@@ -33,10 +55,43 @@ return function( pattern, cldr ) {
...
+					result = getDateTime( "medium", cldr );
+				}
+				else {
+					result = getDateTime( "short", cldr );
+				}
+			}
+			else {
```
rxaviers: If we check for all date and time fields in `flagDate` and `flagTime`, there should be no `else`, right?

rxaviers:
Below is what needs to be done: (from the spec)

> If a client-requested set of fields includes both date and time fields, and if the availableFormats data does not include a dateFormatItem whose skeleton matches the same set of fields, then the request should be handled as follows:

> 1. Divide the request into a date fields part and a time fields part

Can you point me to your code where is this being performed?

> 2. For each part, find the matching dateFormatItem

Can you point me to your code where is this being performed?

> 3. Combine the patterns for the two dateFormatItems using the appropriate dateTimeFormat pattern, determined as follows from the requested date fields:
>  - If the requested date fields include wide month (MMMM, LLLL) and weekday name of any length (e.g. E, EEEE, c, cccc), use <dateTimeFormatLength type="full">
>  - Otherwise, if the requested date fields include wide month, use <dateTimeFormatLength type="long">
>  - Otherwise, if the requested date fields include abbreviated month (MMM, LLL), use <dateTimeFormatLength type="medium">
>  - Otherwise use <dateTimeFormatLength type="short">

rxaviers:
Beyond the above comments, your code needs to follow our coding styles http://contribute.jquery.org/style-guide/js/.

src/date/expand-pattern.js
```diff
@@ -1,3 +1,24 @@
...
+			]),
+			cldr.main([
+				"dates/calendars/gregorian/dateFormats",
+				type
+			])
+		]);
+	}
+	return result;
+}
+
```
rxaviers: 1. By placing this function outside the `define` scope, it's outside the module and therefore won't built correctly.
1. Given you have created this function, why aren't you using it on `case "datetime" in pattern:` below? You have duplicate block of code.

ManrajGrover:
@rxaviers sir,

What I have inferred from the issue is this:
> Divide the request into a date fields part and a time fields part

Here there is no need to do this since the date and time part can be checked in both A and B ways as mentioned by you above and we won't be requiring individual parts as the 4 cases below don't need them. If they are present individually, they have already been handled before.
> For each part, find the matching dateFormatItem

Again since we have to return `"full"` , `"long"`, "`medium`", `"short"` , I see no need to find raw patterns.
> Combine the patterns for the two dateFormatItems using the appropriate dateTimeFormat pattern, determined as follows from the requested date fields

Here we need to return `"full"` , `"long"`, "`medium`", `"short"` according to cases.

This is what I meant in the rough code I mentioned [here](https://github.com/jquery/globalize/issues/272#issuecomment-81966916) 

Now I feel I had inferred it wrong. What I now feel needs to be done is this:
> Divide the request into a date fields part and a time fields part

Here extract the date and time part by parsing and checking for 'M', 'h' etc.
> For each part, find the matching dateFormatItem

Check if these exist individually in the `availableFormats`. What if it doesn't? Throw an error, right?
> Combine the patterns for the two dateFormatItems using the appropriate dateTimeFormat pattern, determined as follows from the requested date fields

Case 1. If date component has `"(MMMM, LLLL)"` and `"(E, EEEE, c, cccc)"` , we use rawPattern of `"full"`. By use we mean return the `"full"` or we need to arrange `"MMMM"` and everything in component as in rawPattern of `"full"` etc, right? Need a suggestion how to do this, if it indeed is to be done.
....(same for next cases)

> Beyond the above comments, your code needs to follow our coding styles http://contribute.jquery.org/style-guide/js/.

Okay sir, going through it.

ManrajGrover:
@rxaviers sir,

May I know if new inference is correct or if there is any changes required.

ManrajGrover:
@rxaviers sir, 

If there is any change in what I mentioned please let me know or give me a yes. so that I start implementing it. Thank you.

rxaviers:
@ManrajGrover, I'm not convinced your interpretation is correct. What pattern do you get for the skeleton `"MMME"`? Do you get `"LLL, ccc"`?

ManrajGrover:
@rxaviers sir,

No. But I guess I have started to see your point now and what the algorithm actually says.

>If a client-requested set of fields includes both date and time fields, and if the availableFormats data does not include a dateFormatItem whose skeleton matches the same set of fields, then the request should be handled as follows:
>
>    1. Divide the request into a date fields part and a time fields part

We need to take out the date and time from the skeleton provided.
>    1. For each part, find the matching dateFormatItem

We match the dateSkeleton and timeSkeleton to the `availableFormats`
>    1. Combine the patterns for the two dateFormatItems using the appropriate dateTimeFormat pattern, determined as follows from the requested date fields:
>        1. If the requested date fields include wide month (MMMM, LLLL) and weekday name of any length (e.g. E, EEEE, c, cccc), use `<dateTimeFormatLength type="full">`

Sir here we will check dateSkeleton and see if it contains `MMMM` and if yes then we will use `"full"` dateTime skeleton, convert it to raw pattern and then map the raw pattern of dateSkeleton to dateTime raw pattern and delete which all unmapped entities from `result` before returning it.
>        1. Otherwise, if the requested date fields include wide month, use `<dateTimeFormatLength type="long">`

Same here with dateTime raw pattern `"long"`.
>        1. Otherwise, if the requested date fields include abbreviated month (MMM, LLL), use `<dateTimeFormatLength type="medium">`

Same here with dateTime raw pattern `"medium"`.
>        1. Otherwise use `<dateTimeFormatLength type="short">`

Same here with dateTime raw pattern `"short"`.

Sorry for taking time to understand this. I hope this is correct now.

rxaviers:
> We need to take out the date and time from the skeleton provided.

Correct. :+1:

> We match the dateSkeleton and timeSkeleton to the `availableFormats`

Correct. :+1:

> Sir here we will check dateSkeleton and see if it contains MMMM and if yes then we will use "full" dateTime skeleton, convert it to raw pattern and then map the raw pattern of dateSkeleton to dateTime raw pattern and delete which all unmapped entities from result before returning it.
> ...
> Same here with dateTime raw pattern "long".
> ...
> Same here with dateTime raw pattern "medium".
> ...
> Same here with dateTime raw pattern "short".

I don't think this is correct. I understood that depending of the fields present in the dateSkeleton or timeSkeleton you should use a different form of  "glue". For example, if the skeleton has "MMMM", you must use `<dateTimeFormatLength type="full">`, which is `{1} 'at' {0}`. You will replace `{1}` with the date pattern and `{0}` with the time pattern according to http://www.unicode.org/reports/tr35/tr35-dates.html#dateTimeFormat.

> Sorry for taking time to understand this. I hope this is correct now.

No problem. Thanks for the resilience.

ManrajGrover:
@rxaviers sir,

Ohkay sir! I was thinking of giving full flexibility to user to use any `skeleton` he/she wants and then mapping accordingly to give the result(This would require some brainstorming as well)

I am currently working on it but one thing that is confusing is `availableFormats` does not contain any `skeleton` for `MMMM`. I will add this case once this is cleared. Till then, I will code rest of the things.

ManrajGrover:
@rxaviers sir,

I have re-coded the whole feature and tested on few cases and got positive results. I have also taken care of putting `G` to the end of the result in case of `medium` and `short` for good result. I have tried to follow good coding style given [here](http://contribute.jquery.org/style-guide/js/) but don't know if I am missing anything or not. 

Also please clear the `MMMM` case as they are not available under `availableFormats` given [here](http://www.unicode.org/repos/cldr-aux/json/26/main/en/ca-gregorian.json).

I have not included `L` and `c` cases as they don't occur in `skeleton` but occur in raw pattern.

Please review the code. Thank you.

jzaefferer:
@ManrajGrover regarding GSoC (I missed your IRC PM), please submit your proposal on the official site, we can review it there.

ManrajGrover:
@jzaefferer sir,

No problem sir. Sure sir. Thank you. :)

src/date/expand-pattern.js
```diff
@@ -33,10 +55,65 @@ return function( pattern, cldr ) {
...
+			}
+			if ( /[GyQMd]/.test( skeleton ) ) {
+				flagDate = true;
+			}
+			if ( /Ed/g.test( skeleton ) ) {
+				flagDate = true;
+			} else if ( /Eh/g.test( skeleton ) || /EH/g.test( skeleton ) ||
+					/E/g.test( skeleton ) ) {
+				flagTime = true;
+			}
```
rxaviers: Why are you testing these various combinations of `E`? `E` means day of week. Therefore, it's a date pattern. `d` means day, therefore it's a date pattern. `h` or `H` means hours, therefore it's a time pattern.

src/date/expand-pattern.js
```diff
@@ -33,10 +55,65 @@ return function( pattern, cldr ) {
...
+				if ( /MMMM/g.test( dateSkeleton ) && /[E]/.test( dateSkeleton ) ) {
+					result = getDateTime( "full", dateSkeleton, timeSkeleton, cldr );
+				} else if ( /MMMM/g.test( dateSkeleton ) ) {
+					result = getDateTime( "long", dateSkeleton, timeSkeleton, cldr );
+				} else if ( /MMM/g.test( dateSkeleton ) ) {
+					result = getDateTime( "medium", dateSkeleton, timeSkeleton, cldr );
+					if ( result.indexOf( "G" ) > -1 ) {
+						result = result.replace( "G", "" );
+						result += " G";
+					}
```
rxaviers: I don't get it why you're testing G here individually?

rxaviers:
> I have re-coded the whole feature and tested on few cases and got positive results. I have also taken care of putting `G` to the end of the result in case of `medium` and `short` for putting `AD` at the end. I have tried to follow good coding style given [here](http://contribute.jquery.org/style-guide/js/) but don't know if I am missing anything or not.

- I see that no unit or functional tests have been added. Therefore, I don't know what tests you are referring to.
- I still see old reviews that haven't been addressed. For example, https://github.com/jquery/globalize/pull/426/files#r26713245.
- I've added more questions above.

> Also please clear the `MMMM` case as they are not available under `availableFormats` given [here](http://www.unicode.org/repos/cldr-aux/json/26/main/en/ca-gregorian.json).

As told in https://github.com/jquery/globalize/issues/272#issuecomment-83159576, this is a different issue (#271) and it's going to be handled separately.

> I have not included `L` and `c` cases as they don't occur in `skeleton` but occur in raw pattern.

Why they don't occur? Could you point me, please, to the excerpt of the specifications that says so?

To give an example, this is what I expect to be done:

On [2.6.2 Elements availableFormats, appendItems](http://www.unicode.org/reports/tr35/tr35-dates.html#availableFormats_appendItems), there is:

> The id attribute is a so-called "skeleton", containing only field information, and in a canonical order. Examples are "yMMMM" for year + full month, or "MMMd" for abbreviated month + day. In particular:
>  - The fields are from the Date Field Symbol Table in Date Format Patterns.
>  - The canonical order is from top to bottom in that table; that is, "yM" not "My".
>  - Only one field of each type is allowed; that is, "Hh" is not valid.
>  - The 'a' field is not allowed in the skeleton.

The patterns of the table are `GyYuUrQqMLlwWdDFgEecahHKkjJmsSAzZOvVXx`. Although, `j` and `J` are explicitly saying they cannot be used in skeletons. Plus the above items says `a` is not allowed in the skeleton as well. Therefore, the allowed patterns of a skeleton are `GyYuUrQqMLlwWdDFgEechHKkmsSAzZOvVXx`.

rxaviers:
@ManrajGrover, right now I'm not sure if I need to assist you on JavaScript or understanding CLDR. My suggestion: let's make sure we're in sync with CLDR first. So, skip all my questions above. Let's focus on the below.

What's the pattern of the skeletons below?
 1. `"dhms"` (for the `en` locale).
 1. `"MMMMEdhm"` (for the `de` locale).
 1. `"MMMMEdhm"` (for the `en` locale).
 1. `"GyMMMEdhms"` (for the `en` locale).
 1. `"GyMMMEdhms"` (for the `en` locale).
 1. `"Ems"` (for the `en` locale).
 1. `"yQQQhm"` (for the `en` locale).
 1. `"GyQQQhm"` (for the `en` locale).

Below, I answer the first three ones. I am waiting for your answer of the others.
 1. `"dhms"` skeleton should return the `"d, h:mm a"` pattern for the `en` locale.
 2. `"MMMMEdhm"` skeleton should return the `"E, d. MMMM `um`  h:mm a"` pattern for the `de` locale.
 3. `"MMMMEdhm"` skeleton should return an error of missing pattern for the `en` locale.

--- 

Explanation for 1.

> 1. Divide the request into a date fields part and a time fields part

- date fields = `d`;
- time fields = `hms`;

> 2. For each part, find the matching dateFormatItem

- `d` => `d`;
- `hms` => `h:mm:ss a`;

> 3. Combine the patterns for the two dateFormatItems using the appropriate dateTimeFormat pattern, determined as follows from the requested date fields:
>  - If the requested date fields include wide month (MMMM, LLLL) and weekday name of any length (e.g. E, EEEE, c, cccc), use \<dateTimeFormatLength type="full"\>
>  - Otherwise, if the requested date fields include wide month, use \<dateTimeFormatLength type="long"\>
>  - Otherwise, if the requested date fields include abbreviated month (MMM, LLL), use \<dateTimeFormatLength type="medium"\>
>  - Otherwise use \<dateTimeFormatLength type="short"\>

Using the short pattern (i.e., `{1}, {0}`) to compose the final pattern: `d, h:mm:ss a` for the `en` locale.

--- 

Explanation for 2.

> 1. Divide the request into a date fields part and a time fields part

- date fields = `MMMMEd`;
- time fields = `hm`;

> 2. For each part, find the matching dateFormatItem

- `MMMMEd` => `E, d. MMMM`;
- `hm` => `h:mm a`;

> 3. Combine the patterns for the two dateFormatItems using the appropriate dateTimeFormat pattern, determined as follows from the requested date fields:
>  - If the requested date fields include wide month (MMMM, LLLL) and weekday name of any length (e.g. E, EEEE, c, cccc), use \<dateTimeFormatLength type="full"\>
>  - Otherwise, if the requested date fields include wide month, use \<dateTimeFormatLength type="long"\>

Using the full pattern (i.e., `{1} `um` {0}`) to compose the final pattern: `"E, d. MMMM `um`  h:mm a"` for the `de` locale.

--- 

Explanation for 3.

> 1. Divide the request into a date fields part and a time fields part

- date fields = `MMMMEd`;
- time fields = `hm`;

> 2. For each part, find the matching dateFormatItem

There's no dateFormatItem for `MMMMEd` on the `en` locale. Therefore, return missing pattern error.

ManrajGrover:
@rxaviers sir,

There are reasons for those. I will explain them. Sure sir, agreed. Lets do CLDR first.

4."GyMMMEdhms" (for the en locale): 

Explanation for 4.

> Divide the request into a date fields part and a time fields part

- date fields = GyMMMEd;
- time fields = hms;

> For each part, find the matching dateFormatItem

- GyMMMEd => E, MMM d, y G;
- hms => h:mm:ss a;

> 1. Combine the patterns for the two dateFormatItems using the appropriate dateTimeFormat pattern, determined as follows from the requested date fields:
>    - If the requested date fields include wide month (MMMM, LLLL) and weekday name of any length (e.g. E, EEEE, c, cccc), use `<dateTimeFormatLength type="full">`
>    - Otherwise, if the requested date fields include wide month, use `<dateTimeFormatLength type="long">`
>    - Otherwise, if the requested date fields include abbreviated month (MMM, LLL), use `<dateTimeFormatLength type="medium">`

According to algorithm, we use the medium pattern (i.e., `{1}, {0}`) to compose the final pattern: `E, MMM d, y G, h:mm:ss a` for the `en` locale. According to me, it should be "E, MMM d, y , h:mm:ss G" because `AD` should come at the end. If its not required, I will remove the test.

5."GyMMMEdhms" (for the en locale)

Explanation for 5.

> Divide the request into a date fields part and a time fields part

- date fields = GyMMMEd;
- time fields = hms;

> For each part, find the matching dateFormatItem

- GyMMMEd => E, MMM d, y G;
- hms => h:mm:ss a;

> 1. Combine the patterns for the two dateFormatItems using the appropriate dateTimeFormat pattern, determined as follows from the requested date fields:
>    - If the requested date fields include wide month (MMMM, LLLL) and weekday name of any length (e.g. E, EEEE, c, cccc), use `<dateTimeFormatLength type="full">`
>    - Otherwise, if the requested date fields include wide month, use `<dateTimeFormatLength type="long">`
>    - Otherwise, if the requested date fields include abbreviated month (MMM, LLL), use `<dateTimeFormatLength type="medium">`

According to algorithm, we use the medium pattern (i.e., `{1}, {0}`) to compose the final pattern: `E, MMM d, y G, h:mm:ss a` for the `en` locale. According to me, it should be "E, MMM d, y , h:mm:ss G" because `AD` should come at the end. If its not required, I will remove the test.

6."Ems" (for the en locale)

Explanation for 6.

> Divide the request into a date fields part and a time fields part

- date fields = E;
- time fields = ms;

> For each part, find the matching dateFormatItem

- E => ccc;
- ms => mm:ss;

> 1. Combine the patterns for the two dateFormatItems using the appropriate dateTimeFormat pattern, determined as follows from the requested date fields:
>    - If the requested date fields include wide month (MMMM, LLLL) and weekday name of any length (e.g. E, EEEE, c, cccc), use `<dateTimeFormatLength type="full">`
>    - Otherwise, if the requested date fields include wide month, use `<dateTimeFormatLength type="long">`
>    - Otherwise, if the requested date fields include abbreviated month (MMM, LLL), use `<dateTimeFormatLength type="medium">`
>    - Otherwise use `<dateTimeFormatLength type="short">`

According to algorithm, we use the short pattern (i.e., `{1}, {0}`) to compose the final pattern: `ccc, mm:ss` for the `en` locale. I see the mistake here.

7."yQQQhm" (for the en locale)

Explanation for 7.

> Divide the request into a date fields part and a time fields part

- date fields = yQQQ;
- time fields = ms;

> For each part, find the matching dateFormatItem

- yQQQ => QQQ y;
- ms => mm:ss;

> 1. Combine the patterns for the two dateFormatItems using the appropriate dateTimeFormat pattern, determined as follows from the requested date fields:
>    - If the requested date fields include wide month (MMMM, LLLL) and weekday name of any length (e.g. E, EEEE, c, cccc), use `<dateTimeFormatLength type="full">`
>    - Otherwise, if the requested date fields include wide month, use `<dateTimeFormatLength type="long">`
>    - Otherwise, if the requested date fields include abbreviated month (MMM, LLL), use `<dateTimeFormatLength type="medium">`
>    - Otherwise use `<dateTimeFormatLength type="short">`

According to algorithm, we use the short pattern (i.e., `{1}, {0}`) to compose the final pattern: `QQQ y, mm:ss` for the `en` locale.

8."GyQQQhm" (for the en locale)

Explanation for 8.

> Divide the request into a date fields part and a time fields part

- date fields = GyQQQ;
- time fields = hm;

`GyQQQ` does not exist as skeleton hence would return an error.

Sir, could you please answer some more?

1. `GyMMMEdEHms` (for the en locale)
1. `GyMMMEdE` (for the en locale)
1. `GyMMME` (for the en locale)
1. `MMME` (for the en locale)

rxaviers:
Great! That's it. But, note the comment that follows.

> According to me, it should be "E, MMM d, y , h:mm:ss G" because `AD` should come at the end. If its not required, I will remove the test.

If we really believe this change should be made, we should contact CLDR and have
them to change their data. We should not modify that ourselves. So, we should
follow the algorithm.

rxaviers:
> Sir, could you please answer some more?

Sure.

> 1. `GyMMMEdEHms` (for the en locale)

This skeleton is incorrect for two reasons.
1. There should be no duplicate `E`.
1. `E` should never appear after `e`, `c`, `a`, `h`, `H`, `K`, `k`, `j`, `J`, `m`, `s`, `S`, `A`, `z`, `Z`, `O`, `v`, `V`, `X`, or `x`.

See skeleton definition at http://www.unicode.org/reports/tr35/tr35-dates.html#availableFormats_appendItems.

> `GyMMMEdE` (for the en locale)

Same as above.

> `GyMMME` (for the en locale)

> 1. Divide the request into a date fields part and a time fields part

- date fields = `GyMMME`;
- time fields = ``;

> 2. For each part, find the matching dateFormatItem

There's no dateFormatItem for `GyMMME` on the `en` locale. Therefore, return missing pattern error.

> `MMME` (for the en locale)

Analogous to the above.

rxaviers:
@ManrajGrover, given this clarification phase, I think the next step is to jump to the JS code. Just let me know when I can make another review round? Note my comments and just let me know in any questions.

src/date/expand-pattern.js
ManrajGrover: @rxaviers sir,

Sir here it is throwing error `E_MISSING_CLDR` error from `cldr.main` itself and not returning `undefined` for skeletons containing both date and time part. Hence we first have to check if skeleton entered has dateSkeleton or timeSkeleton or both before getting result. I am attaching a screenshot that will explain better.
untitled213

src/date/expand-pattern.js
ManrajGrover: @rxaviers sir,

Fixing.

ManrajGrover:
@rxaviers sir,

Okay sir. I will try to point out one or two things which made me add these changes. One of them has been mentioned by you. Some of them I will comment on the diffs itself.

>Great! That's it. But, note the comment that follows.
>
>>According to me, it should be "E, MMM d, y , h:mm:ss G" because AD should come at the end. If its not required, I will remove the test.

>If we really believe this change should be made, we should contact CLDR and have them to change their data. We should not modify that ourselves. So, we should follow the algorithm.

Sure sir. I will remove this part for now.

>@ManrajGrover, I'm not convinced your interpretation is correct. What pattern do you get for the skeleton `"MMME"`? Do you get `"LLL, ccc"`?

Sir here you gave me an example that for `"MMME"` it should be `"LLL, ccc"` but now you mentioned it should return an error. This is why I was testing for E to either be in Time or in Date. Fixing it now.

>@ManrajGrover, given this clarification phase, I think the next step is to jump to the JS code. Just let me know when I can make another review round? Note my comments and just let me know in any questions.

Sure sir.

rxaviers:
> Sir here you gave me an example that for `"MMME"` it should be `"LLL, ccc"` but now you mentioned it should return an error. This is why I was testing for E to either be in Time or in Date. Fixing it now.

My bad. Sorry about it.

rxaviers:
Having `MMME` to return an error should be the correct behavior.

src/date/expand-pattern.js
rxaviers: > Hence we first have to check if skeleton entered has dateSkeleton or timeSkeleton or both before getting result. I am attaching a screenshot that will explain better.

Nope. There may be skeletons containing both date and time parts. So, we need to follow what's in the specification, which says "if the availableFormats data does not include a dateFormatItem whose skeleton matches the same set of fields" that we should look for date and time separately.

We need to find a way to technically do that. `cldr.main` itself doesn't throw `E_MISSING_CLDR` error. The error is thrown by our validation. We could skip that path from the validation. See https://github.com/jquery/globalize/blob/master/src/date.js#L63.

rxaviers: For example, otherwise you wouldn't get `EHms`.

rxaviers:
@ManrajGrover, the current implementation is incorrect. I assume we both are in sync with that and you will fix it and let me know. Please, just let me know otherwise. Thank you

src/date/expand-pattern.js
ManrajGrover: @rxaviers sir,

>There may be skeletons containing both date and time parts. So, we need to follow what's in the specification, which says "if the availableFormats data does not include a dateFormatItem whose skeleton matches the same set of fields" that we should look for date and time separately.

Yeah. Some cases will get missed through this method. 

>We could skip that path from the validation. See https://github.com/jquery/globalize/blob/master/src/date.js#L63.

By skipping the path you mean adding the path for `availableFormats` here?
https://github.com/jquery/globalize/blob/master/src/date.js#L29

rxaviers: > By skipping the path you mean adding the path for availableFormats here?
https://github.com/jquery/globalize/blob/master/src/date.js#L29

Correct.

ManrajGrover:
@rxaviers sir,

I feel everything is covered now. Please review the code and let me know if I am missing anything. Thank you.

src/date.js
```diff
@@ -29,7 +29,8 @@ function validateRequiredCldr( path, value ) {
 		skip: [
 			/dates\/calendars\/gregorian\/days\/.*\/short/,
 			/supplemental\/timeData\/(?!001)/,
-			/supplemental\/weekData\/(?!001)/
+			/supplemental\/weekData\/(?!001)/,
+			/dates\/calendars\/gregorian\/dateTimeFormats\/availableFormats/
```
rxaviers: Style: please sort using alphabetical order, i.e., as first line.

src/date/expand-pattern.js
rxaviers: Please, see my comment above (1) and (2) items.

src/date/expand-pattern.js
```diff
@@ -33,10 +55,30 @@ return function( pattern, cldr ) {
...
+				for ( i = 0 ; i < skeleton.length ; i++ ) {
+					if ( /[hHms]/.test( skeleton[i] ) ) {
+						timeSkeleton += skeleton[i];
+					} else if ( /[GyQMEd]/.test( skeleton[i] ) ) {
+						dateSkeleton += skeleton[i];
+					}
+				}
+				if ( ( /MMMM/g.test( dateSkeleton ) || /LLLL/g.test( dateSkeleton ) ) &&
+						/[Ec]/.test( dateSkeleton ) ) {
+					result = getDateTime( "full", dateSkeleton, timeSkeleton, cldr);
```
rxaviers: Style: add space:

```diff
- cldr);
+ cldr );
```

src/date/expand-pattern.js
```diff
@@ -33,10 +55,30 @@ return function( pattern, cldr ) {
 
 	switch ( true ) {
 		case "skeleton" in pattern:
+			skeleton = pattern.skeleton;
 			result = cldr.main([
 				"dates/calendars/gregorian/dateTimeFormats/availableFormats",
-				pattern.skeleton
+				skeleton
 			]);
+			if ( result === undefined ) {
```
rxaviers: This could be safely simplified to `if ( !result ) {` or alternatively:

```js
if ( result ) {
   break;
}
// code below...
```

src/date/expand-pattern.js
```diff
@@ -33,10 +55,30 @@ return function( pattern, cldr ) {
...
+				for ( i = 0 ; i < skeleton.length ; i++ ) {
+					if ( /[hHms]/.test( skeleton[i] ) ) {
+						timeSkeleton += skeleton[i];
+					} else if ( /[GyQMEd]/.test( skeleton[i] ) ) {
+						dateSkeleton += skeleton[i];
+					}
```
rxaviers: Why `hHms` and `HyQMEd` only? You have ignored what I've said in https://github.com/jquery/globalize/pull/426#issuecomment-85226376 "allowed patterns of a skeleton are `GyYuUrQqMLlwWdDFgEechHKkmsSAzZOvVXx`"

src/date/expand-pattern.js
```diff
@@ -33,10 +55,30 @@ return function( pattern, cldr ) {
...
+				}
```
rxaviers: Instead of the whole for loop, you could:

```js
timeSkeleton = skeleton.replace( /[^hHms]/g, "" );
```

Analogous thing for dateSkeleton.

src/date/expand-pattern.js
```diff
@@ -33,10 +55,30 @@ return function( pattern, cldr ) {
...
+				if ( ( /MMMM/g.test( dateSkeleton ) || /LLLL/g.test( dateSkeleton ) ) &&
```
rxaviers: Alternatively:

```js
if ( /(MMMM|LLLL).*[Ec]/.test( dateSkeleton ) ) {
```

src/date/expand-pattern.js
rxaviers: Note this approach may lead to undesired results. For example, take this invalid skeleton `MhMhMh` as an input. We would wrongly have timeSkeleton `hhh` and dateSkeleton `MMM`.

One way to fix that is using the below approach:

```js
timeSkeleton = skeleton.split( /[^hHms]/ ).slice( -1 )[ 0 ];
```

PS: Note I'm using `hHms`, but it should obviously be extended with all the correct time fields.

PS2: For dateSkeleton, `slice` is not needed.

rxaviers: Note this approach may lead to undesired results. For example, take this invalid skeleton `MhMhMh` as an input. We would wrongly have timeSkeleton `hhh` and dateSkeleton `MMM`.

One way to fix that is using the below approach:

```js
timeSkeleton = skeleton.split( /[^hHms]/ ).slice( -1 )[ 0 ];
```

PS: Note I'm using `hHms`, but it should obviously be extended with all the correct time fields.

PS2: For dateSkeleton, `slice` is not needed.

src/date/expand-pattern.js
ManrajGrover: @rxaviers sir,

Sir, there is difference in path used in `cldr.main` of blocks of code mentioned. If wanted, we can add a `if` block for `case "datetime" in pattern:` and check for `"datetime" in pattern:` or `"skeleton" in pattern:` to handle both in one function. Should I add?

ManrajGrover:
@rxaviers sir,

Made the changes. Please review the fixes. Thank you.

ManrajGrover:
Hi @rxaviers sir,

Please review the code and let me know the changes. Sorry if you're busy and I am disturbing. Thank you.

src/date.js
```diff
@@ -28,6 +28,7 @@ function validateRequiredCldr( path, value ) {
 	validateCldr( path, value, {
 		skip: [
 			/dates\/calendars\/gregorian\/days\/.*\/short/,
+			/dates\/calendars\/gregorian\/dateTimeFormats\/availableFormats/,
 			/supplemental\/timeData\/(?!001)/,
 			/supplemental\/weekData\/(?!001)/
```
rxaviers: Alphabetical order is:
```js
/dates\/calendars\/gregorian\/dateTimeFormats\/availableFormats/,
/dates\/calendars\/gregorian\/days\/.*\/short/,
/supplemental\/timeData\/(?!001)/,
/supplemental\/weekData\/(?!001)/
```

src/date/expand-pattern.js
```diff
@@ -24,8 +24,30 @@ define([
...
+					timeSkeleton
+				]),
+				cldr.main([
+					"dates/calendars/gregorian/dateTimeFormats/availableFormats",
+					dateSkeleton
+				])
+			]);
+		}
+		return result;
+	};
```
rxaviers: On https://github.com/jquery/globalize/pull/426#discussion_r26713245, I have asked you:

> 2. Given you have created this function, why aren't you using it on case "datetime" in pattern: below? You have duplicate block of code.

You have replied:

> Sir, there is difference in path used in `cldr.main` of blocks of code mentioned. If wanted, we can add a `if` block for `case "datetime" in pattern:` and check for `"datetime" in pattern:` or `"skeleton" in pattern:` to handle both in one function. Should I add?

The only difference is the paths:
- https://github.com/ManrajGrover/globalize/blob/fix-272-handleDateTime/src/date/expand-pattern.js#L40-L41 vs https://github.com/ManrajGrover/globalize/blob/fix-272-handleDateTime/src/date/expand-pattern.js#L95-L96
- https://github.com/ManrajGrover/globalize/blob/fix-272-handleDateTime/src/date/expand-pattern.js#L44-L45 vs https://github.com/ManrajGrover/globalize/blob/fix-272-handleDateTime/src/date/expand-pattern.js#L99-L100

The rest has the same logic. Can you reuse the function somehow (for example, by using additional arguments)?

src/date.js
ManrajGrover: Oopsie. Fixing.

src/date/expand-pattern.js
ManrajGrover: Yes sir we can. But that would also lead to increase in size of these lines which I thought has more overhead:

https://github.com/ManrajGrover/globalize/blob/fix-272-handleDateTime/src/date/expand-pattern.js#L67
https://github.com/ManrajGrover/globalize/blob/fix-272-handleDateTime/src/date/expand-pattern.js#L69
https://github.com/ManrajGrover/globalize/blob/fix-272-handleDateTime/src/date/expand-pattern.js#L71
https://github.com/ManrajGrover/globalize/blob/fix-272-handleDateTime/src/date/expand-pattern.js#L73

Surely we can add two more arguments `datePath` and `timePath` if that is what is wanted.

rxaviers: > Surely we can add two more arguments `datePath` and `timePath` if that is what is wanted.

You can use the if's to determine `type`. Then, call `getDateTime` once.

ManrajGrover: @rxaviers sir,

Fixed it. Please review the code.

src/date/expand-pattern.js
```diff
@@ -24,8 +24,30 @@ define([
  * - { datetime: "full" } returns "EEEE, MMMM d, y 'at' h:mm:ss a zzzz";
  * - { pattern: "dd/mm" } returns "dd/mm";
  */
+
 return function( pattern, cldr ) {
-	var result;
+	var result, getDateTime, skeleton, dateSkeleton = "", timeSkeleton = "", type;
```
rxaviers:
```diff
-       var result, getDateTime, skeleton, dateSkeleton = "", timeSkeleton = "",
+       var getDateTime, result, skeleton, type,
+               dateSkeleton = "",
+               timeSkeleton = "";
```

Note that variables that assign initial value are placed in their own lines. Note the alphabetical order.

src/date/expand-pattern.js
```diff
@@ -33,10 +55,27 @@ return function( pattern, cldr ) {
...
+			skeleton = pattern.skeleton;
 			result = cldr.main([
 				"dates/calendars/gregorian/dateTimeFormats/availableFormats",
-				pattern.skeleton
+				skeleton
 			]);
+			if ( !result ) {
+				timeSkeleton = skeleton.split( /[^hHKkmsSAzZOvVXx]/ ).slice( -1 )[ 0 ];
+				dateSkeleton = skeleton.split( /[^GyYuUrQqMLlwWdDFgEec]/ )[ 0 ];
+				if (  /(MMMM|LLLL).*[Ec]/.test( dateSkeleton ) ) {
```
rxaviers: Extra space on `  /(MM...`

src/date/expand-pattern.js
```diff
@@ -33,10 +55,27 @@ return function( pattern, cldr ) {
...
+				} else if ( /MMMM/g.test( dateSkeleton ) ) {
+					type = "long";
+				} else if ( /MMM/g.test( dateSkeleton ) || /LLL/g.test( dateSkeleton ) ) {
+					type = "medium";
+				} else {
+					type = "short";
+				}
+				result = getDateTime( type, dateSkeleton, timeSkeleton,
+						"dates/calendars/gregorian/dateTimeFormats/availableFormats",
+						"dates/calendars/gregorian/dateTimeFormats/availableFormats", cldr );
```
rxaviers: There's something I don't like `getDateTime()`. It's not too intuitive to understand it.

Our goal here is to merge two pattern strings (e.g., `MMM y G` and `HH:mm:ss`) using a merge-pattern (e.g., `{1}, {0}`). For example, producing `MMM y G, HH:mm:ss`.

I suggest changing the code according to:

```diff
-	var result, getDateTime, skeleton, dateSkeleton = "", timeSkeleton = "", type;
+	var dateSkeleton, result, skeleton, timeSkeleton, type;
 
-	getDateTime = function( type, dateSkeleton, timeSkeleton, datePath, timePath, cldr ) {
-		var result;
-		result = cldr.main([
-			"dates/calendars/gregorian/dateTimeFormats",
-			type
-		]);
-		if ( result ) {
-			result = formatMessage( result, [
-				cldr.main([
-					timePath,
-					timeSkeleton
-				]),
-				cldr.main([
-					datePath,
-					dateSkeleton
-				])
-			]);
-		}
-		return result;
-	};
+	function combineDateTime( type, datePattern, timePattern ) {
+		return formatMessage(
+			cldr.main([
+				"dates/calendars/gregorian/dateTimeFormats",
+				type
+			]),
+			[ datePattern, timePattern ]
+		);
+	}
 
 	if ( typeof pattern === "string" ) {
 		pattern = { skeleton: pattern };
@@ -72,9 +61,16 @@ return function( pattern, cldr ) {
 				} else {
 					type = "short";
 				}
-				result = getDateTime( type, dateSkeleton, timeSkeleton,
+				result = combineDateTime( type,
+					cldr.main([
 						"dates/calendars/gregorian/dateTimeFormats/availableFormats",
-						"dates/calendars/gregorian/dateTimeFormats/availableFormats", cldr );
+						dateSkeleton
+					]),
+					cldr.main([
+						"dates/calendars/gregorian/dateTimeFormats/availableFormats",
+						timeSkeleton
+					])
+				);
 			}
 			break;
 
@@ -88,9 +84,16 @@ return function( pattern, cldr ) {
 			break;
 
 		case "datetime" in pattern:
-			result = getDateTime( pattern.datetime, pattern.datetime, pattern.datetime,
+			result = combineDateTime( pattern.datetime,
+				cldr.main([
 					"dates/calendars/gregorian/dateFormats",
-					"dates/calendars/gregorian/timeFormats", cldr );
+					pattern.datetime
+				]),
+				cldr.main([
+					"dates/calendars/gregorian/timeFormats",
+					pattern.datetime
+				])
+			);
 			break;
 
 		case "pattern" in pattern:
```

rxaviers:
Implementation is working, great! Now, it's time to fix the coding style issues for which I've left comments inline and consider replacing `getDateTime` for `combineDateTime` detailed in my inline comments.

src/date/expand-pattern.js
ManrajGrover: Missed it. Fixed.

src/date/expand-pattern.js
ManrajGrover: Sir, I tried the code suggested by you and it worked great but there is a flaw in `combineDateTime()` function. I feel it should be `[ timePattern, datePattern ]` and not `[ datePattern, timePattern ]` comparing the results of previous code with new implementation. Hence committing this change as well. Let me know if its wrong.

ManrajGrover:
@rxaviers sir,

I have replaced `getDateTime` for `combineDateTime` but found a bug in the function and have commented issue and fix for it. Let me know if I missed anything.

rxaviers:
@ManrajGrover, yeap `date` and `time` were inverted there. Your fix is correct.

rxaviers:
Landed thanks

ManrajGrover:
@rxaviers sir,

Great! Thank you for helping me throughout the fix. I feel we should document the new feature now and add few examples for the same.

rxaviers:
Sounds good. Feel free to send a PR with such updates. Thanks

PR #431, opened 2015-04-10
Build: Update to QUnit 1.18, load it via AMD as well

jzaefferer:
Extract requirejs configuration into separate file. Load synchronosly
for the grunt-contrib-qunit bridge to be able to use it as well.

Increase the waitTimeout to make it work on with browserstack-runner - need to verify it now works reliable, could then also merge #381.

Also fix a typo in a module name.


rxaviers:
:+1: thanks

Issue #398, opened 2015-02-17
Runtime code (smaller and quicker) for production


rxaviers:
PR #436

**Goal**

On client side, performance is also about how fast our page loads. So, size matters. How to get the smallest/leanest bundle for production?

Suppose we need a plural function for English. That is what we need: a function that given a number, outputs the plural form. But, in order to get that, we need the English plural rules (from CLDR) and the library code to parse that rules in order to generate this simple function `function( count ) { return count == 1 ? "one" : "other"; }`. In the end, this tiny thing is what matters. What if we could precompile that at build time and deploy only this tiny function.

Our goal is to allow our formatters/parsers to output the precompiled function. So, by managing that at build time, we can remove the need for most of the library at runtime.

**General ideas**

- All formatters/parsers should expose a `.toString()` function to output itself as a clean string representation. Below, I'm referring to this function as runtime-formatters/runtime-parsers.
- runtime-formatters/runtime-parsers should not depend on CLDR data. But, it should depend on pre-processed properties only.
- runtime-formatters/runtime-parsers can reference other functions. Those functions should be packed and distributed as the runtime library for each module. For example, `globalize/date.runtime.js`, `globalize/number.runtime.js`, etc.

**References**
- http://handlebarsjs.com/
- [messageformat.js](https://github.com/SlexAxton/messageformat.js/blob/722ebabd4d64adb61025d45aa87c5c6a438e63f3/README.md#build-time-compilation)

jzaefferer:
Handlebars also does this pretty well, providing a separate runtime to load with compiled templates. In combination with more-or-less simple requirejs plugins, the whole things can be automated pretty well.

rxaviers:
Yeap, handlebars is actually the original motivator for messageformat.js as well. I've included its link as another reference. Thanks

srk12345:
hi i'm shahrukh ur idea is good and sounding.Plz tell me how can i contribute in ur project.

rxaviers:
Hi @srk12345, thanks for your interest on contributing to Globalize. A good initial reading is https://github.com/jquery/globalize/blob/master/CONTRIBUTING.md. Then, get yourself familiarized with the contributing workflow by fixing simpler issues, e.g., #393. Then, get familiarized with our source code. If you have any question just let us know.

srk12345:
how can i start my work on plural function for english

rxaviers:
[...]

Issue #413, opened 2015-03-11
Update coding style according to new style guide updates


rxaviers:
http://contribute.jquery.org/style-guide/js/

jzaefferer:
The `onevar` rule is gone, some spacing exceptions were removed. We're planning to use esformatter to update the entire codebase at once (at least for the spacing rules). Need to land major PRs first, to avoid tons of conflicts.

PR #436, opened 2015-04-23
Runtime support (smaller and quicker code for production)

rxaviers:
Fix for #398

Related created projects:
- https://github.com/jquery-support/globalize-compiler
- https://github.com/rxaviers/globalize-webpack-plugin

TODO:
- [x] Docs
- [x] Tests
 - [x] Check if runtimeBind is being properly set.
 -  Check if the compiled code runs on the supported browsers #482.


dpolivy:
Any timeline for when this (cache support) may be incorporated?

rxaviers:
> Any timeline for when this (cache support) may be incorporated?

I've moved the Cache support for its own PR #470.

rxaviers:
@jquery/globalize, necessary documentation has been included and cache support has been moved into its own PR. So, this PR is (finally) ready for review.

README.md
```diff
@@ -136,15 +136,15 @@ information on its usage.
```
rxaviers: Content team (cc @arthurvr @agcolom @kswedberg who has helped reviewing past doc changes), please just let me know if you have thoughts for improvements. Thanks in advance.

rxaviers: PS: Save your time by ignoring the file renames. :P

jzaefferer:
@rxaviers I don't have the bandwidth to review 12k LOC inbetween. We could do that together some time, let me know when you're available.

Maybe you can pull out the examples changes into a separate PR (like you kinda did with the cache)? That would make reviewing this a lot easier...

README.md
```diff
@@ -248,11 +250,84 @@ functionalities, etc.
...
-- [Hello World (Node.js + npm)](examples/node-npm/).
-- [Hello World (plain JavaScript)](examples/plain-javascript/).
+- [Hello World (AMD + bower)](examples/hello-world-amd-bower/).
+- [Hello World (Node.js + npm)](examples/hello-world-node-npm/).
+- [Hello World (plain JavaScript)](examples/hello-world-plain-javascript/).
 
 
+### Performance
+
+When formatting or parsing, there's actually a two step process: (a) the
```
kswedberg: two-step process (hyphenate two-step)

README.md
```diff
@@ -248,11 +250,84 @@ functionalities, etc.
...
+formatter (or parser) creation and (b) its execution, where creation takes
```
kswedberg: maybe make first mention of *creation* and *execution* italics or bold?

README.md
```diff
@@ -248,11 +250,84 @@ functionalities, etc.
...
+considerably more time (more expensive) than execution. The difference is an
```
kswedberg: Instead of "considerably" here, you could just write "an order of magnitude" and then remove the following sentence.

README.md
```diff
@@ -248,11 +250,84 @@ functionalities, etc.
...
+```
+
+As a rule of thumb for optimal performance, cache your formatters and parsers.
+For example: (a) on iterations, generate them outside the loop and reuse while
+looping; (b) on server applications, generate them in advance and execute when
+requests arrive.
+
+### Compilation and the Runtime modules
+
+You really should take advantage of compiling your formatters and/or parsers
```
kswedberg: You can remove "You really should" here and begin the sentence with "Take advantage". Or, you could start with "You should" and simply remove the "really."

README.md
```diff
@@ -248,11 +250,84 @@ functionalities, etc.
...
+<script src="globalize-runtime/number.js"></script>
+<script src="globalize-runtime/...js"></script>
+<script src="my-compiled-formatters-and-parsers.js"></script>
+<script>
+// Use Globalize formatters and parsers.
+</script>
+```
+
+Your compiled formatters and parsers allow you to skip a big part of the library
+and also allow you to skip loading CLDR data. Because, they have already been
```
kswedberg: Combine the two sentences: 
"Your compiled formatters and parsers allow you to skip a big part of the library and skip loading CLDR data because they…"

examples/basic-globalize-compiler/README.md
```diff
@@ -0,0 +1,26 @@
+# Basic Globalize Compiler example
+
+On this example, we assume you know the Globalize usage basics. So, it focuses on
```
kswedberg: Maybe change these two sentences to:
"This example focuses on the Globalize Compiler and the Globalize runtime modules. It assumes knowledge of Globalize usage basics."

examples/basic-globalize-compiler/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+1. Point your browser at `./development.html`. Note that the formatters are
+created dynamically. Therefore, Cldrjs and CLDR data are required.
+1. Understand the demo by reading the source code. We have comments there for
+you.
+
+### Production mode
+
+1. Install `globalize-compiler` by running `npm install`.
+1. Compile the application formatters by running `npm run build`. See
```
kswedberg: Aside: Is this the recommended way of creating a numbered list in markdown—using `1.` for every item? I've never seen it done this way.

arthurvr: We're doing it inconsistent all over the place. One day we need a markdown style guide but I'd say it's alright for now.

examples/basic-globalize-compiler/app.js
```diff
@@ -0,0 +1,33 @@
+var like, number;
+
+// Set default locale as "en".
+Globalize.locale( "en" );
+
+// Use Globalize to format dates.
+document.getElementById( "date" ).innerHTML = Globalize.formatDate( new Date(), {
+	datetime: "medium"
+});
```
arthurvr: `} );`, or don't you use the new rules yet?

examples/basic-globalize-compiler/README.md
```diff
@@ -0,0 +1,26 @@
+# Basic Globalize Compiler example
+
+On this example, we assume you know the Globalize usage basics. So, it focuses on
+the Globalize Compiler and on the Globalize runtime modules.
```
arthurvr: Generally we don't really do the hard wrapping, just write this out on the same line. Or is this how you're doing it everywhere in this repo?

README.md
rxaviers: Like "where creation takes an order of magnitude more time (more expensive) than execution"? Just double checking.

examples/basic-globalize-compiler/README.md
rxaviers: Any other way? Except by using incrementing 1., 2., 3. ourselves for easier maintenance? :)

arthurvr: `0.` maybe makes it more clear that we're using this trick. Though I think it's alright as-is.

examples/basic-globalize-compiler/app.js
rxaviers: I plan this update to be done on #413

arthurvr: :+1:

examples/basic-globalize-compiler/README.md
rxaviers: Yeap, all the documentation throughout this repo has been using hard wrapping similarly to JS styles ("Lines should be no longer than 80 characters, and must not exceed 100 (counting tabs as 4 spaces). There are 2 exceptions, both allowing the line to exceed 100 characters").

Do we have a rule for that somewhere? I'm just curious.

examples/basic-globalize-compiler/README.md
rxaviers: Good point. I've started gathering documentation style updates on #471.

examples/basic-globalize-compiler/README.md
arthurvr: > Do we have a rule for that somewhere? I'm just curious.

Hard wrapping in markdown is sortof annoying as you can't change a sentence without having to change a bunch of unrelated lines. Though, we don't have a style guide about this, so it boils down to a preference.

Once we had a tiny [discussion](https://github.com/jquery/learn.jquery.com/issues/285#issue-11579167) about it but that's specific for the learn site.

rxaviers:
@kswedberg thanks. All your observations so far have been fixed by 34697a4

rxaviers:
> Maybe you can pull out the examples changes into a separate PR (like you kinda did with the cache)? That would make reviewing this a lot easier...

@jzaefferer, but the example is tightly coupled with this change. In fact, I think starting with the example is the best way to start reviewing this change. Because, it shows what can be accomplished using the runtime modules.

rxaviers:
> @rxaviers I don't have the bandwidth to review 12k LOC inbetween. We could do that together some time, let me know when you're available.

@jzaefferer, I understand and sure I'm pinging you privately to schedule.

robbbz:
> @jzaefferer, I understand and sure I'm pinging you privately to schedule.

Can I point you at the fact that the word "privately" doesn't really fit into an open-source mindset?

rxaviers:
@robbbz thanks for your observation. Privately there means: not to bug others unnecessarily. I'm open to suggestions on how we can improve anything here.

kswedberg:
Nice work, @rxaviers !

README.md
```diff
@@ -248,11 +250,83 @@ functionalities, etc.
...
+looping; (b) on server applications, generate them in advance and execute when
+requests arrive.
+
+### Compilation and the Runtime modules
+
+Take advantage of compiling your formatters and/or parsers during build time
+when deploying to production. It's much faster than generating them in real-time
+and it's also much smaller (i.e., better loading performance).
+
+To illustrate, a Globalize application looks like:
```
jzaefferer: The blanks here seem problematic to me. I think full examples with no blanks would be better. You can leave out irrelevant parts, but the "..." parts here are all pretty important.

README.md
```diff
@@ -248,11 +250,83 @@ functionalities, etc.
...
+</script>
+```
+
+A precompiled Globalize application using the runtime modules looks like:
+
+```html
+<script src="globalize-runtime.js"></script>
+<script src="globalize-runtime/number.js"></script>
+<script src="globalize-runtime/...js"></script>
+<script src="my-compiled-formatters-and-parsers.js"></script>
```
jzaefferer: This, to me, begs the question: Why isn't this just a single file?

src/common/runtime-bind.js
```diff
@@ -0,0 +1,23 @@
+define([
+	"./runtime-key"
+], function( runtimeKey ) {
+
+return function( args, cldr, fn, runtimeArgs ) {
+	var fnName = /^function\s+([\w\$]+)\s*\(/.exec( fn.toString() )[ 1 ],
```
jzaefferer: Can't you just access `fn.name`?

src/build/intro-currency-runtime.js
```diff
@@ -0,0 +1,43 @@
...
+		// AMD
+		define([
+			"../globalize-runtime",
+			"./number"
+		], factory );
+	} else if ( typeof exports === "object" ) {
+
+		// Node, CommonJS
+		module.exports = factory(
+			require( "globalize/dist/globalize-runtime" ),
```
jzaefferer: Why can't you use relative paths here (like for AMD above)?

src/build/intro-currency-runtime.js
```diff
@@ -0,0 +1,43 @@
+/**
```
jzaefferer: Can you have a single template for all the intro file headers and generate them at build time? There's a lot of duplication here.

src/common/create-error/plural-module-presence.js
```diff
@@ -0,0 +1,9 @@
+define([
+	"../create-error"
+], function( createError ) {
+
+return function() {
+	return createError( "E_MISSING_PLURAL_MODULE", "Plural module not loaded." );
```
jzaefferer: As far as I can tell this is used exactly once. I don't think that's worth the extra file.

src/common/runtime-bind.js
```diff
@@ -0,0 +1,23 @@
...
+		locale = cldr.locale;
+
+	fn.runtimeKey = runtimeKey( fnName, locale, args );
+
+	// Stringify arguments.
+	args = JSON.stringify( args ).slice( 1, -1 );
```
jzaefferer: Is `args` guaranteed to never contain a cyclical structure? Afaik that's the only reason for `JSON.stringify` to throw a `TypeError`

jzaefferer: Also overwriting a variable with a different type is usually a bad idea. Might want to use a different variable or inline below.

src/common/runtime-bind.js
```diff
@@ -0,0 +1,23 @@
+define([
+	"./runtime-key"
+], function( runtimeKey ) {
+
+return function( args, cldr, fn, runtimeArgs ) {
+	var fnName = /^function\s+([\w\$]+)\s*\(/.exec( fn.toString() )[ 1 ],
+		locale = cldr.locale;
```
jzaefferer: Let's stop using onevar?

src/common/runtime-bind.js
```diff
@@ -0,0 +1,23 @@
...
+
+	fn.runtimeKey = runtimeKey( fnName, locale, args );
```
jzaefferer: This also calls `JSON.stringify(args)`, do we really need to do that twice?

src/common/runtime-key.js
```diff
@@ -0,0 +1,12 @@
+define([
+	"../util/string/hash"
+], function( stringHash ) {
+
+return function( fnName, locale, args ) {
+	var hash;
+	args = [].slice.call( args, 0 );
```
jzaefferer: Why is this needed?

src/currency.js
```diff
@@ -1,14 +1,14 @@
...
 	"./common/validate/cldr",
 	"./common/validate/default-locale",
 	"./common/validate/parameter-presence",
 	"./common/validate/parameter-type/currency",
 	"./common/validate/parameter-type/number",
 	"./common/validate/parameter-type/plain-object",
-	"./common/validate/plural-module-presence",
 	"./currency/code-properties",
-	"./currency/name-format",
+	"./currency/formatter/fn",
```
jzaefferer: `fn` is not a good module or function name. Its called as a function, so it much be a function, so what is it doing?

test/functional/currency/currency-formatter.js
```diff
@@ -45,7 +45,9 @@ QUnit.module( ".currencyFormatter( currency [, options] )", {
 		});
 		Globalize.locale( "en" );
 	},
-	teardown: util.resetCldrContent
+	teardown: function() {
+		util.resetCldrContent();
```
jzaefferer: Why this change? The function doesn't seem to care in which scope it is called. At least keep it out of this PR? 17 less files changed is worth it.

jzaefferer:
What are you plans for test automation for the runtime? That seems pretty crucial for any future changes to any module.

src/number.js
```diff
@@ -19,11 +20,11 @@ define([
 
 	"cldr/event",
 	"cldr/supplemental"
-], function( Globalize, createErrorUnsupportedFeature, validateCldr, validateDefaultLocale,
-	validateParameterPresence, validateParameterRange, validateParameterTypeNumber,
-	validateParameterTypePlainObject, validateParameterTypeString, numberFormat,
-	numberFormatProperties, numberNumberingSystem, numberParse, numberParseProperties,
-	numberPattern, numberSymbol, stringPad ) {
+], function( Globalize, createErrorUnsupportedFeature, runtimeBind, validateCldr,
```
jzaefferer: I still hate these epic argument lists. #445 might not be perfect, but at least it improves this. Btw. would still be interesting to know what the size difference on top of runtime would be...

src/number.js
```diff
@@ -37,14 +38,16 @@ define([
  */
 Globalize.numberFormatter =
 Globalize.prototype.numberFormatter = function( options ) {
-	var cldr, maximumFractionDigits, maximumSignificantDigits, minimumFractionDigits,
-		minimumIntegerDigits, minimumSignificantDigits, pattern, properties;
+	var args, cldr, maximumFractionDigits, maximumSignificantDigits, minimumFractionDigits,
```
jzaefferer: This epic var list also looks dysfunctional. As a starting point to improve: All the min/maxXxx vars are used in one place, just once or maybe twice. All for validating `properties`. That would be trivial to extract into a separate function (in the same file).

Also without onvar most of these could be defined where they are actually used, making refactorings like the one described above much easier.

src/number/format-properties.js
```diff
@@ -45,12 +45,18 @@ return function( pattern, cldr, options ) {
 	negativePrefix = negativeProperties[ 0 ];
 	negativeSuffix = negativeProperties[ 10 ];
 
+	// Have runtime code to refer to numberRound() instead of including it explicitly.
```
jzaefferer: Why is this necessary?

jzaefferer:
Related to the lack of tests, there's so many details, like the `numberRound` optimization that will be trivial to miss when working on any future features or bugfixes.

src/number.js
jzaefferer: Btw. having worked with ES2015 modules (with babel for transpiling) for 3 months, AMD define is pretty ugly in comparison... Looking forward to switching here.

src/common/runtime-bind.js
```diff
@@ -0,0 +1,23 @@
...
+
+	// Stringify arguments.
+	args = JSON.stringify( args ).slice( 1, -1 );
+
+	fn.generatorString = function() {
+		return "Globalize(\"" + locale + "\")." + fnName + "(" + args + ")";
+	};
+
+	fn.runtimeArgs = runtimeArgs;
```
jzaefferer: Related to the lack of tests: This property is never accessed anywhere (in this PR). I guess a code coverage tool wouldn't notice either, but its actually pretty crucial.

src/currency/formatter/fn.js
```diff
@@ -0,0 +1,32 @@
...
+	"../../common/validate/parameter-presence",
+	"../../common/validate/parameter-type/number",
+	"../name-format"
+], function( validateParameterPresence, validateParameterTypeNumber, currencyNameFormat ) {
+
+return function( numberFormatter, pluralGenerator, properties ) {
+	var fn;
+
+	// Return formatter when style is "code" or "name".
+	if ( pluralGenerator && properties ) {
```
jzaefferer: This looks strange. Wouldn't it be better to split these two paths into two files and "include" just the one that's actually gonna be used?

src/date-runtime.js
```diff
@@ -0,0 +1,55 @@
...
+Globalize._dateFormatterFn = dateFormatterFn;
+Globalize._dateParserFn = dateParserFn;
+Globalize._dateFormat = dateFormat;
+Globalize._dateParser = dateParse;
+Globalize._dateTokenizer = dateTokenizer;
+Globalize._validateParameterTypeDate = validateParameterTypeDate;
+
+Globalize.dateFormatter =
+Globalize.prototype.dateFormatter = function( options ) {
+	options = options || { skeleton: "yMd" };
```
jzaefferer: Is there any way to avoid duplicating this default (also defined in `src/date.js`)? Otherwise fixing #466 will have to change both, and whatever future changes we'll do have the same problem.

src/message.js
```diff
@@ -90,15 +82,18 @@ Globalize.prototype.messageFormatter = function( path ) {
 	}
 	validateMessageType( path, message );
 
-	formatter = MessageFormatInit( this, cldr ).compile( message );
+	isPluralModulePresent = this.plural !== undefined;
```
jzaefferer: This is used exactly once. Doesn't justify a variable.

src/message/formatter-runtime-bind.js
```diff
@@ -0,0 +1,45 @@
...
+
+		if ( /plural\(/.test( output ) ) {
+			args.plural = "messageFormat.plural";
+		}
+
+		if ( /select\(/.test( output ) ) {
+			args.select = "messageFormat.select";
+		}
+
+		output.replace( /pluralFuncs(\[([^\]]+)\]|\.([a-zA-Z]+))/, function( match ) {
```
jzaefferer: I have no idea what this is doing. This specifically, the whole file in general.

src/number-runtime.js
```diff
@@ -0,0 +1,57 @@
+define([
```
jzaefferer: All the `src/*-runtime.js` follow the same patterns. The `format/parseX` methods validate parameters, the others map arguments through `runtimeKey`. Is there any way to get rid of these and generate them in the build or with the compiler?

src/util/string/hash.js
```diff
@@ -0,0 +1,12 @@
+define(function() {
+
+// Based on http://stackoverflow.com/questions/7616461/generate-a-hash-from-string-in-javascript-jquery
+return function( str ) {
+	return [].reduce.call( str, function( hash, i ) {
```
jzaefferer: As far as I can tell, this will get called pretty often, so its probably worth optimizing. You can avoid lots of function calls why using a regular loop, ala https://github.com/jquery/qunit/blob/29ca2e87c420f6dfdef793b034a67cff9a666df5/src/test.js#L432-L435

jzaefferer:
I just took a peek at https://github.com/jquery-support/globalize-compiler/blob/b0.0.1/src/compile.js - that looks like it has to know awfully much about Globalize implementation details. Any change on Globalize might require a change on the compiler. That seems way too brittle.

jzaefferer:
I followed the instructions in `examples/basic-globalize-example` and got this error:

```sh
$ npm run build

> basic-globalize-compiler@ build /Users/jza/dev/globalize/examples/basic-globalize-compiler
> globalize-compiler -l en -c cldr.json -m messages.json -o compiled-formatters.js app.js

/Users/jza/dev/globalize/examples/basic-globalize-compiler/node_modules/globalize-compiler/src/compile.js:168
		throw new Error( "No formatters or parsers has been provided" );
		      ^
Error: No formatters or parsers has been provided
    at compiler (/Users/jza/dev/globalize/examples/basic-globalize-compiler/node_modules/globalize-compiler/src/compile.js:168:9)
    at Object.compileExtracts (/Users/jza/dev/globalize/examples/basic-globalize-compiler/node_modules/globalize-compiler/src/compile-extracts.js:58:9)
    at Object.<anonymous> (/Users/jza/dev/globalize/examples/basic-globalize-compiler/node_modules/globalize-compiler/bin/globalize-compiler.js:82:28)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.<anonymous> (/Users/jza/dev/globalize/examples/basic-globalize-compiler/node_modules/globalize-compiler/bin/globalize-compiler:3:1)
```

README.md
rxaviers:
```html
<!-- Globalize Runtime -->
<script src="globalize-runtime.js"></script>
<script src="globalize-runtime/number.js"></script>
<script src="globalize-runtime/...js"></script>

<!-- My precompiled formatters and parsers -->
<script src="my-compiled-formatters-and-parsers.js"></script>
```

README.md
rxaviers: There is a full example link below. Do you suggest to completely remove this pseudo-example from here or adjust it?

README.md
rxaviers: Is this additional comments enough? Would you expect something different?

src/common/runtime-bind.js
rxaviers: Ah! Thanks. Much better.

src/build/intro-currency-runtime.js
rxaviers: It comes with a price of fancier build tools. I'm open for PRs implementing this change.

src/build/intro-currency-runtime.js
rxaviers: Thanks. Good point.

src/common/create-error/plural-module-presence.js
rxaviers: As an organizational point of view for source files, all custom errors are placed there.

As a performance point of view for distribution files, this function becomes an inlined variable. So, there's no extra function wrapping.

src/common/runtime-bind.js
rxaviers: > Is `args` guaranteed to never contain a cyclical structure? Afaik that's the only reason for `JSON.stringify` to throw a `TypeError`

`args` is always formatters/parsers `arguments` after all type validations have been performed. It's not supposed to be cyclic. But, some arguments are potential candidates to be cyclic, for example if user passes an option with cyclic keys (despite the fact this would be really awkward) this is not verified.

Do you think it's worth handling that in advance, so JSON.stringify doesn't fail? Or it would be ok failing given the unexpected use case?

src/common/runtime-bind.js
rxaviers: #413

src/common/runtime-key.js
rxaviers: At some point of the implementation, formatters/parsers `arguments` as is were been passed. But, now an Array clone is always the case. So, this isn't needed anymore indeed. Thanks

src/common/runtime-bind.js
rxaviers: Good point. Passing as argument.

src/common/runtime-key.js
rxaviers: Actually, this is still needed because of https://github.com/jquery/globalize/blob/2baa3317e2726cc6d5f9c0f8895958f78ab9aa34/src/message-runtime.js#L14. But, I'm adapting this call, so we keep runtimeKey simpler.

src/common/runtime-bind.js
rxaviers: Actually, it's passed twice by the normal modules. But, it's passed once by the runtime modules (e.g., https://github.com/jquery/globalize/blob/2baa3317e2726cc6d5f9c0f8895958f78ab9aa34/src/date-runtime.js#L28).

I'm considering to handle both worlds with:
```diff
diff --git a/src/common/runtime-key.js b/src/common/runtime-key.js
index 45b64e5..4d95f06 100644
--- a/src/common/runtime-key.js
+++ b/src/common/runtime-key.js
@@ -3,8 +3,9 @@ define([
 ], function( stringHash ) {
 
 return function( fnName, locale, args ) {
-       var hash;
-       hash = stringHash( fnName + locale + JSON.stringify( args ) );
+       var argsStr, hash;
+       argsStr = typeof args === "string" ? args : JSON.stringify( args );
+       hash = stringHash( fnName + locale + argsStr );
        return hash > 0 ? "a" + hash : "b" + Math.abs( hash );
 };
```

rxaviers: 9f3fe6f24d6e50c732c24422d686c7fdc84b547c

src/common/runtime-key.js
rxaviers: a16e8309eb9600b923bbecf34cd43febe95ee275

src/build/intro-currency-runtime.js
rxaviers: a0fdc94bcc7957aa36da74f0eac22c3a8961a217

src/common/runtime-bind.js
rxaviers: a03a92db431a0d2c9e42ed3b24c8654ce490e200

README.md
rxaviers: 6d637fd9c2bb9ef6074c2e77ab6bff15103f1c59

src/currency.js
rxaviers: This is supposed to be the returned function by the generator. Take for example `numberFormatter`, it returns a formatter, that's what this function is about.

I am going to name it `./currency/formatter` only and make the same for all other analogous. Feel free to suggest other alternatives.

src/common/runtime-bind.js
rxaviers: I'm finally using this:
```diff
diff --git a/src/common/runtime-key.js b/src/common/runtime-key.js
index 45b64e5..c1513aa 100644
--- a/src/common/runtime-key.js
+++ b/src/common/runtime-key.js
@@ -2,9 +2,10 @@ define([
        "../util/string/hash"
 ], function( stringHash ) {
 
-return function( fnName, locale, args ) {
+return function( fnName, locale, args, argsStr ) {
        var hash;
-       hash = stringHash( fnName + locale + JSON.stringify( args ) );
+       argsStr = argsStr || JSON.stringify( args );
+       hash = stringHash( fnName + locale + argsStr );
        return hash > 0 ? "a" + hash : "b" + Math.abs( hash );
 };
```

rxaviers: d08dba82bf621c367ce79be26e744b0f6aad9db4

src/currency.js
rxaviers: EDIT: This is supposed to be the returned function by the generator. Take for example `numberFormatter`, it returns a formatter, that's what this function is about.

I am going to name it `./currency/formatter-fn` and make the same for all other analogous. Feel free to suggest other alternatives.

rxaviers: 02e97563c03aae6d4b35fe338c76f95aa35d5c4c

test/functional/currency/currency-formatter.js
rxaviers: Indeed, this is a preparation for a Cache change (https://github.com/jquery/globalize/commit/76f16d605ba904510f9aed9b58699d1c78cd22b0#diff-7762adfa36f42583983ea0b59eb78b71R49). So, this should live in #470.

rxaviers: fda340aac8bd2770309eabc795a9c10c43ff25b7

src/number.js
rxaviers: Me too.

src/number.js
rxaviers: A PR would help to see the your proposal.

src/number/format-properties.js
rxaviers: To generate stuff like:

```js
Globalize.a1378886668 = numberFormatterFn(["",,1,,,,,,,,"","0","-0","-","",numberRound(),"∞","NaN",{".":".",",":",","%":"%","+":"+","-":"-","E":"E","‰":"‰"},]);
```

Instead of

```js
Globalize.a1378886668 = numberFormatterFn(["",,1,,,,,,,,"","0","-0","-","",function ( value, incrementOrExp ) {
		var exp, increment;

		value = +value;

		// If the value is not a number, return NaN.
		if ( isNaN( value ) ) {
			return NaN;
		}

		// Exponent given.
		if ( typeof incrementOrExp === "object" && incrementOrExp.exponent ) {
			exp = +incrementOrExp.exponent;
			increment = 1;

			if ( exp === 0 ) {
				return method( value );
			}

			// If the exp is not an integer, return NaN.
			if ( !( typeof exp === "number" && exp % 1 === 0 ) ) {
				return NaN;
			}

		// Increment given.
		} else {
			increment = +incrementOrExp || 1;

			if ( increment === 1 ) {
				return method( value );
			}

			// If the increment is not a number, return NaN.
			if ( isNaN( increment ) ) {
				return NaN;
			}

			increment = increment.toExponential().split( "e" );
			exp = +increment[ 1 ];
			increment = +increment[ 0 ];
		}

		// Shift & Round
		value = value.toString().split( "e" );
		value[ 0 ] = +value[ 0 ] / increment;
		value[ 1 ] = value[ 1 ] ? ( +value[ 1 ] - exp ) : -exp;
		value = method( +( value[ 0 ] + "e" + value[ 1 ] ) );

		// Shift back
		value = value.toString().split( "e" );
		value[ 0 ] = +value[ 0 ] * increment;
		value[ 1 ] = value[ 1 ] ? ( +value[ 1 ] + exp ) : exp;
		return +( value[ 0 ] + "e" + value[ 1 ] );
	},"∞","NaN",{".":".",",":",","%":"%","+":"+","-":"-","E":"E","‰":"‰"},]);
```

rxaviers:
> Related to the lack of tests, there's so many details, like the `numberRound` optimization that will be trivial to miss when working on any future features or bugfixes.

Tests should cover it.

src/common/runtime-bind.js
rxaviers: Tests could ensure it's there.

rxaviers: EDIT: Tests should ensure it's there.

src/currency/formatter/fn.js
rxaviers: Yes, it would. Note this function is returned by the formatter generator. So, doing it properly takes a little more than just splitting this into two files. Or you'd end up still having the `if` condition in the parent function and having actually no much gain.

rxaviers: Further PRs could be made to improve this.

src/date-runtime.js
rxaviers: Perhaps the solution proposed in #466 will avoid duplicating this.

src/message.js
rxaviers: 598be18a743dcbdcc3814ff3f0886cabd805594d

src/message/formatter-runtime-bind.js
rxaviers: It adjusts SlexAxton/messageformat.js compiled variables with Globalize variables.

rxaviers: 9ac65d0

Feel free to add comments in case this still needs clarification.

src/number-runtime.js
rxaviers: The runtime library and the compiled formatters/parsers are two distinct things and should remain that way. You should load the runtime library once and could load multiple compiled formatters/parsers sequentially.

src/util/string/hash.js
rxaviers: The faster this code, the better. I'm curious how faster your suggested code is compared to this one. The difference is using `for` instead of the ES5 `reduce`.

rxaviers:
> I just took a peek at https://github.com/jquery-support/globalize-compiler/blob/b0.0.1/src/compile.js - that looks like it has to know awfully much about Globalize implementation details. Any change on Globalize might require a change on the compiler. That seems way too brittle.

GlobalizeCompiler is tight coupled with Globalize. Perhaps having GlobalizeCompiler to live here (Globalize repository) could help by allowing changes to touch both places in the same time. This https://github.com/facebook/react keeps some [packages](https://github.com/facebook/react/tree/master/packages) the same way.

rxaviers:
> I followed the instructions in examples/basic-globalize-example and got this error:

The example expects the installed Globalize to support runtime, which will happen only after this lands and gets published. For now, it needs manual installation.

src/number-runtime.js
rxaviers: I'm open to discuss this further...

rxaviers:
> GlobalizeCompiler is tight coupled with Globalize. Perhaps having GlobalizeCompiler to live here (Globalize repository) could help by allowing changes to touch both places at the same time. For comparison, [facebook/react](https://github.com/facebook/react) keeps some [packages](https://github.com/facebook/react/tree/ecb34de5747a6d4578dc4be920a033db9f81370f/packages) the same way.

Thinking more about it, I believe both projects (globalize and globalize-compiler) should be kept in their separate repos. The comparison I gave for react packages makes more sense for globalize and globalize-runtime than for globalize and globalize-compiler. For now, their integration will be tested here by (a) runtime bindings unit tests and (b) app examples that uses the compiler (examples/basic-globalize-compiler and the upcoming examples/app-npm-webpack #464 and examples/app-amd-grunt #465); and their (globalize-compiler) by its own unit tests.

rxaviers:
@jzaefferer thanks a lot for all your comments. Most of them have resulted in fix commits. For the others, I have added a reply.

If anyone has any comments, feel free to make them. We should be good to get this landed.

README.md
rxaviers: I've completely remove the pseudo-example and now we have "To illustrate, see our [Basic Globalize Compiler example][]." f99fa8803802b2c88f1a593989911715816f486d

test/util.js
```diff
@@ -169,6 +169,18 @@ return {
...
+	assertRuntimeBind: function( assert, formatter, runtimeKey, generatorString, runtimeArgsFn ) {
+		assert.ok( formatter );
+		assert.equal( typeof formatter, "function" );
+		assert.ok( "runtimeKey" in formatter );
+		assert.equal( formatter.runtimeKey, runtimeKey );
+		assert.ok( "generatorString" in formatter );
+		assert.equal( typeof formatter.generatorString, "function" );
+		assert.equal( formatter.generatorString(), generatorString );
+		assert.ok( "runtimeArgs" in formatter );
+		runtimeArgsFn( formatter.runtimeArgs );
```
jzaefferer: It looks like this avoids a temporary variable for the `assertRuntimeBind` called. I recommend refactoring that to actually use the temporary variable, since it'll hold the object that is actually being tested. Then drop the `runtimeArgsFn` argument and just call whatever other assertion with `formatter.runtimeArgs`. That removes a level of nesting in several tests along with a noop callback argument in many places.

test/util.js
```diff
@@ -169,6 +169,18 @@ return {
...
 	 */
+	assertRuntimeBind: function( assert, formatter, runtimeKey, generatorString, runtimeArgsFn ) {
+		assert.ok( formatter );
+		assert.equal( typeof formatter, "function" );
+		assert.ok( "runtimeKey" in formatter );
+		assert.equal( formatter.runtimeKey, runtimeKey );
+		assert.ok( "generatorString" in formatter );
+		assert.equal( typeof formatter.generatorString, "function" );
+		assert.equal( formatter.generatorString(), generatorString );
+		assert.ok( "runtimeArgs" in formatter );
```
jzaefferer: I'd also drop all `ok` assertions that are made redundant by follow-up assertions, so all 4 here.

test/util.js
```diff
@@ -169,6 +169,18 @@ return {
 	/**
 	 * Etc
 	 */
+	assertRuntimeBind: function( assert, formatter, runtimeKey, generatorString, runtimeArgsFn ) {
+		assert.ok( formatter );
+		assert.equal( typeof formatter, "function" );
+		assert.ok( "runtimeKey" in formatter );
+		assert.equal( formatter.runtimeKey, runtimeKey );
```
jzaefferer: This assertion seems brittle, in that it'll have to be updated whenever the formatter/parser somehow changes form, independent of the underlying algorithm to generate the key. Does it really makes sense to hardcode the expected key in so many places?

test/util.js
```diff
@@ -169,6 +169,18 @@ return {
 	/**
 	 * Etc
 	 */
+	assertRuntimeBind: function( assert, formatter, runtimeKey, generatorString, runtimeArgsFn ) {
+		assert.ok( formatter );
+		assert.equal( typeof formatter, "function" );
```
jzaefferer: The `formatter` can also be a "parser". Maybe rename to "actual", to reflect that this is what's being tested?

test/functional/number/number-formatter.js
```diff
@@ -77,4 +77,45 @@ QUnit.test( "should return a formatter", function( assert ) {
...
+		Globalize.numberFormatter(),
+		"b468386326",
+		"Globalize(\"en\").numberFormatter({})",
+		function( runtimeArgs ) {
+			assert.equal( JSON.stringify( runtimeArgs[ 0 ] ), "[\"\",null,1,0,3,null,null," +
+				"null,3,null,\"\",\"#,##0.###\",\"-#,##0.###\",\"-\",\"\",null,\"∞\",\"NaN\",{" +
+				"\".\":\".\",\",\":\",\",\"%\":\"%\",\"+\":\"+\",\"-\":\"-\",\"E\":\"E\",\"‰\":" +
+				"\"‰\"},null]"
+			);
+			assert.ok( "generatorString" in runtimeArgs[ 0 ][ 15 ] );
```
jzaefferer: This is another redundant `ok`, the `equal` below is sufficient.

test/functional/message/message-formatter.js
```diff
@@ -175,4 +175,32 @@ QUnit.test( "should support ICU message format", function( assert ) {
...
+
+	util.assertRuntimeBind(
+		assert,
+		Globalize( "en" ).messageFormatter( "like" ),
+		"b328290139",
+		"Globalize(\"en\").messageFormatter(\"like\")",
+		function( runtimeArgs ) {
+			assert.equal(
+				runtimeArgs[ 0 ].toString(),
+				"(function( number, plural, pluralFuncs ) {\n  return function (d) { return plural(d.count, 1, pluralFuncs.en, { 0: function() { return \"Be the first to like this\";}, 1: function() { return \"You liked this\";}, one: function() { return \"You and \" + d.someone + \" liked this\";}, other: function() { return \"You and \" + number(d.count, 1) + \" others liked this\";} }); }\n})(messageFormat.number, messageFormat.plural, {\"en\": Globalize(\"en\").pluralGenerator()})"
```
jzaefferer: Too much code... a less specific assertion might improve this.

jzaefferer:
I've reviewed the tests, since I hadn't seen those before. Still doesn't fill me with confidence that someone else is going to be able to contribute to this codebase...

I wanted to check the other additional commits, but since everything has the same commit message, its hard to figure out what each change is about. That's a pretty big drawback of using "fixup!" without any other information. I'll dig some more anyway...

jzaefferer:
Actually it was just the last set of example updates that had generic commit messages, the rest had some detail in the body. That helped!

src/util/string/hash.js
jzaefferer: Dunno if there's a difference in practice. Function calls have a price used to be a good rule of thumb. What I find more important here: I find the `for` loop easier to understand than `reduce()`.

src/build/intro-currency-runtime.js
jzaefferer: Generating only the header should be pretty trivial, don't you think? The rest of the files obviously would be much harder to generate...

jzaefferer:
Still far from a full review, but at least covered a good chunk. Glad to see that the PR overall got a lot simpler (less LOC changed, less files). Seems mostly due to changes in the example.

test/util.js
rxaviers: The `ok` assertions are there for improved error message in case these attributes are not present. The following assertions would break instead of allowing QUnit to properly report a failing test.

test/util.js
rxaviers: I'm open to ideas on how to better test this.

rxaviers: Given third library code (i.e., globalize-compiler) tightly depends on it, I've came up with this very strict tests.

test/util.js
rxaviers: Actually, this allows a custom tests to be performed by each calling code. For example: https://github.com/jquery/globalize/pull/436/files#diff-9ee4a6f38e08bc680f4668d25dedaff8R30, https://github.com/jquery/globalize/pull/436/files#diff-3b257e2c8b293a57b0336b413514f034R110. Please, just let me know if I misinterpreted your comment.

test/util.js
rxaviers: 8ae20ba

test/functional/message/message-formatter.js
rxaviers: I'm open to ideas.

rxaviers:
@jzaefferer thanks for all your reviews so far. They have been pretty helpful.

I believe the feature implemented by this PR is a big plus to the library. It's essential and mandatory to allow developers to use Globalize for production in real projects and should land. So, I suggest that we don't postpone this PR any longer. I acknowledge there are polishings to be done. But, we can always work on improvements as next steps.

I am considering to land what we have here, because I believe it's mature enough to go as phase-1. All improvements are welcome as next step. Anyone that discourages this decision, please say so.

Thanks

test/util.js
```diff
@@ -169,6 +169,18 @@ return {
...
+	assertRuntimeBind: function( assert, formatterOrParser, runtimeKey, generatorString, runtimeArgsFn ) {
+		assert.ok( formatterOrParser );
+		assert.equal( typeof formatterOrParser, "function" );
+		assert.ok( "runtimeKey" in formatterOrParser );
+		assert.equal( formatterOrParser.runtimeKey, runtimeKey );
+		assert.ok( "generatorString" in formatterOrParser );
+		assert.equal( typeof formatterOrParser.generatorString, "function" );
+		assert.equal( formatterOrParser.generatorString(), generatorString );
+		assert.ok( "runtimeArgs" in formatterOrParser );
+		runtimeArgsFn( formatterOrParser.runtimeArgs );
```
jzaefferer: > Actually, this allows a custom tests to be performed by each calling code. For example: https://github.com/jquery/globalize/pull/436/files#diff-9ee4a6f38e08bc680f4668d25dedaff8R30, https://github.com/jquery/globalize/pull/436/files#diff-3b257e2c8b293a57b0336b413514f034R110. Please, just let me know if I misinterpreted your comment.

Its called synchronously with an argument passed by the caller. Thats trivial to do without the callback. From test/functional/number/number-formatter.js:

```js
util.assertRuntimeBind(
	assert,
	Globalize.numberFormatter(),
	"b468386326",
	"Globalize(\"en\").numberFormatter({})",
	function( runtimeArgs ) {
		assert.equal( JSON.stringify( runtimeArgs[ 0 ] ), "[\"\",null,1,0,3,null,null," +
			"null,3,null,\"\",\"#,##0.###\",\"-#,##0.###\",\"-\",\"\",null,\"∞\",\"NaN\",{" +
			"\".\":\".\",\",\":\",\",\"%\":\"%\",\"+\":\"+\",\"-\":\"-\",\"E\":\"E\",\"‰\":" +
			"\"‰\"},null]"
		);
		assert.ok( "generatorString" in runtimeArgs[ 0 ][ 15 ] );
		assert.equal( runtimeArgs[ 0 ][ 15 ].generatorString(), "numberRound()" );
	}
);
```

turns to

```js
var formatter = Globalize.numberFormatter()	
util.assertRuntimeBind( assert, formatter, "b468386326", "Globalize(\"en\").numberFormatter({})" );
assert.equal( JSON.stringify( formatter.runtimeArgs[ 0 ] ), "[\"\",null,1,0,3,null,null," +
	"null,3,null,\"\",\"#,##0.###\",\"-#,##0.###\",\"-\",\"\",null,\"∞\",\"NaN\",{" +
	"\".\":\".\",\",\":\",\",\"%\":\"%\",\"+\":\"+\",\"-\":\"-\",\"E\":\"E\",\"‰\":" +
	"\"‰\"},null]"
);
assert.ok( "generatorString" in formatter.runtimeArgs[ 0 ][ 15 ] );
assert.equal( formatter.runtimeArgs[ 0 ][ 15 ].generatorString(), "numberRound()" );
```
And you can drop the noop function arguments and you get simpler stacktraces.

jzaefferer:
GitHub says this branch has conflicts that much be resolved.

test/util.js
rxaviers: :+1: let's get this improved in a subsequent PR. Could you send it please?

rxaviers: #486

PR #445, opened 2015-05-07
WIP: Refactor validation helpers into a single module

### LINKED ISSUE #362: Validation (internal code): evaluate a single entrypoint API
Jörn raised this concern in https://github.com/jquery/globalize/pull/351#discussion_r21466383. Evaluate the pros and cons of this change.

Currently, no final `dist/` have unused validations. So, it's important to determine how this change would affect that and how would it affect its final sizes.

---

src/currency.js @ header

@jzaefferer 

> Have you considered making this a single validate module that exports an object with multiple functions? Would help reduce the number of dependencies you need to require in many places, including these modules itself, since most share the same dependencies."

@rxaviers

> Different methods require different validations. Therefore, the dependency of the validate functions across Globalize (core), date module, number module, plural module, etc are different. 
> 
> How bundling all validations on a single export affect the final dist files? What the gain is?"

@jzaefferer 

> There are 103 effective LOC in common/validate, counting only the returned functions, excluding the define wrappers and blank lines. Grouping those 14 files into a single one with about 120 LOC would simplify dependencies in many places and improve the overall understandibility (I may have made that word up, I hope you catch my meaning).
> 
> It would also give the potential for other refactorings, as refactorings in general do, for example I suspect that some of the other modules in common are only used in the various validation modules.

jzaefferer:
Fixes #362 

I made the functional changes, but can get past jshint:dist. Apparently the build system isn't able to handle this module type. `onBuildWrite` is kinda complicated, so I'm hoping for some help to debug this. Once it actually builds, we should get some numbers from the compare_size task.

I definitely like the raw reduction in LOCs, currently 291 additions and 456 deletions. Less dependencies and factory arguments is nice.


rxaviers:
The outcome of your change is:

| | Minified size | size 𝛥 | Minified+gz size | size 𝛥 |
| --- | --: | --: | --: | --: |
| globalize.min.js | 3906 | 53.18% | 1672 | 33.23% |
| globalize/currency.min.js | 5751 | -3.68% | 2469 | -5.44% |
| globalize/date.min.js | 15243 | 0.07% | 4959 | -0.74% |
| globalize/message.min.js | 15892 | -3.60% | 4920 | -4.84% |
| globalize/number.min.js | 7226 | -0.17% | 2965 | -1.66% |
| globalize/plural.min.js | 5270 | -2.68% | 2231 | -1.76% |
| globalize/relative-time.min.js | 1468 | 0.69% | 697 | -1.55% |

The whole purpose of keeping each validation separately is being able to balance the weight, i.e., only including them when necessary. In practice, the current code will never load a validation that's not needed. After this change, it will.

Having said that, I'm mostly concerned/interested in seem the delta difference in the runtime modules (#398). Because, they use even less validations, therefore the impact will be bigger.

jzaefferer:
What do these numbers mean?

rxaviers:
The size of final modules.

jzaefferer:
Specifically, how can I read the percentages?

rxaviers:
Delta sizes (comparing master with this PR)

jzaefferer:
So the main module gets bigger, the others get smaller? 

Doing the comparison on top of the runtime changes seems reasonable.

rxaviers:
> So the main module gets bigger, the others get smaller? 

yeap

rxaviers:

Issue #320, opened 2014-09-29
Support for non-gregorian calendars


rxaviers:
WIP PR #447

---

CLDR has data for a variety of different calendars (listed in the table below). Globalize 0.x supported some non-gregorian calendars. But, Globalize 1.x hasn't migrated them yet (it supports the gregorian calendar only).

| Calendar | Globalize 0.x | Globalize 1.x | ibm-js/ecma402 | dojo |
| --- | --- | --- | --- | --- |
| buddhist (Thai Buddhist calendar) | x |  | x | x |
| chinese (Traditional Chinese calendar) |  |  |  |  |
| coptic (Coptic calendar) |  |  |  |  |
| dangi (Traditional Korean calendar) |  |  |  |  |
| ethioaa (Ethiopic calendar, Amete Alem (epoch approx. 5493 B.C.E)) |  |  |  |  |
| ethiopic (Ethiopic calendar, Amete Mihret (epoch approx, 8 C.E.)) |  |  |  |  |
| gregory (Gregorian calendar) | x | x | x | x |
| hebrew (Traditional Hebrew calendar) |  |  | x | x |
| indian (Indian calendar) |  |  |  |  |
| islamic (Islamic calendar) | x |  | x | x |
| islamic-umalqura (Islamic calendar, Umm al-Qura) | x |  | x | x |
| islamic-tbla (Islamic calendar, tabular (intercalary years [2,5,7,10,13,16,18,21,24,26,29] - astronomical epoch)) |  |  |  |  |
| islamic-civil (Islamic calendar, tabular (intercalary years [2,5,7,10,13,16,18,21,24,26,29] - civil epoch)) |  |  | x |  |
| islamic-rgsa (Islamic calendar, Saudi Arabia sighting) |  |  |  |  |
| iso8601 (ISO calendar (Gregorian calendar using the ISO 8601 calendar week rules)) |  |  |  |  |
| japanese (Japanese Imperial calendar) |  |  | x |  |
| persian (Persian calendar) |  |  |  | x |
| roc (Republic of China calendar) |  |  | x |  |
| islamicc (Civil (algorithmic) Arabic calendar) |  |  |  |  |

We haven't been requested so far to support any non-gregorian calendar. Therefore, our priority is (1) first to support the previously supported calendars on 0.x, (2) then to support any other non-gregorian calendar.

References:
- [UTS#35 Dates](http://www.unicode.org/reports/tr35/tr35-dates.html).
[...]

PR #446, opened 2015-05-07
Add support for non-Gregorian calendars

dwachss:
As discussed in jquery/globalize#320 , this creates a generalized date class and allows for the use of non-gregorian calendars. This is more of a proof-of concept; only Gregorian dates are implemented but other calendar algorithms could be added to `Globalize.calendars`.

### Some thoughts: 
Gdate should be an independent class; having `calendars` in `Globalize` makes the dependency graph circular (core.js depends on parse.js which depends on core.js for the calendar algorithms).

start-of.js is ugly; it has both the date and Gdate parameters, which must refer to the same date. That ought to be fixed.

I'm getting an error with the functional tests: 
```
    .dateFormatter( pattern ) - should return a formatter
    Actual: null
    Expected: undefined
    Error: E_MISSING_CLDR: Missing required CLDR content
    'main/en/dates/calendars/gregorian/dateTimeFormats/availableFormats/dhms'.
```
But that looks like a problem with the cldr data.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "globalize",
-  "version": "1.0.0-alpha.21",
+  "version": "1.0.0",
   "license": "MIT",
   "ignore": [
     "**/.*",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "globalize",
-  "version": "1.0.0-alpha.21",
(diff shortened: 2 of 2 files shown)


dwachss:
I changed the pull request to use a dedicated branch, rather than `master`. The new one is jquery/globalize#447.

PR #447, opened 2015-05-08
Add support for non-Gregorian calendars

dwachss:
As discussed in jquery/globalize#320 , this creates a generalized date class and allows for the use of non-gregorian calendars. This is more of a proof-of concept; only Gregorian dates are implemented but other calendar algorithms could be added to `Globalize.calendars`.
### Some thoughts:

Gdate should be an independent class; having `calendars` in `Globalize` makes the dependency graph circular (core.js depends on parse.js which depends on core.js for the calendar algorithms).

start-of.js is ugly; it has both the date and Gdate parameters, which must refer to the same date. That ought to be fixed.

I'm getting an error with the functional tests: 

```
    .dateFormatter( pattern ) - should return a formatter
    Actual: null
    Expected: undefined
    Error: E_MISSING_CLDR: Missing required CLDR content
    'main/en/dates/calendars/gregorian/dateTimeFormats/availableFormats/dhms'.
```

But that looks like a problem with the cldr data.

I'm also getting errors from commitplease, but I don't know how to correct the messages on old commits.


dwachss:
The latest commit (@d8f6e02) now completes the grunt tasks without error.

src/core.js
```diff
@@ -28,6 +35,18 @@ function validateLikelySubtags( cldr ) {
...
+	var calendar;
+	validateParameterPresence( locale, "locale" );
+	validateParameterTypeLocale( locale, "locale" );
+	object.cldr = alwaysCldr( locale );
+	calendar = calendarsCalendarForLocale( object.cldr );
+	validateParameterType ( calendar, "calendar",
+		calendar in Globalize.calendars, "a defined calendar system" );
+	object.cldr.attributes.calendar = calendar;
+	validateLikelySubtags( object.cldr );
+}
```
rxaviers: Please, revert this. Let's handle calender in the date module (not in core).

src/date.js
```diff
@@ -63,6 +63,7 @@ Globalize.prototype.dateFormatter = function( options ) {
 	cldr.on( "get", validateRequiredCldr );
 	pattern = dateExpandPattern( options, cldr );
 	properties = dateFormatProperties( pattern, cldr );
+	properties.calendar = Globalize.calendars[cldr.attributes.calendar];
```
rxaviers: All properties must be set in `dateFormatProperties`. Let's set `calendar` in there as well.

rxaviers: Inside `dateFormatProperties`, you won't have a reference for `Globalize` anymore. Instead of using `Globalize.calendars` as a hub for calendars, simply use something independent, e.g., `calendar/calendars.js`.


```js
define([
	"./calendars/Gregorian-date",
	"./calendars/Hebrew-date",
	"./calendars/Islamic-date"
], function(calendarsGregorianDate, calendarsHebrewDate, calendarsIslamicDate) {

return {
  gregorian: calendarsGregorianDate,
  hebrew: calendarsHebrewDate,
  islamic: calendarsIslamicDate
};

});
```

src/date.js
```diff
@@ -104,6 +105,7 @@ Globalize.prototype.dateParser = function( options ) {
 	pattern = dateExpandPattern( options, cldr );
 	tokenizerProperties = dateTokenizerProperties( pattern, cldr );
 	parseProperties = dateParseProperties( cldr );
+	parseProperties.calendar = cldr.attributes.calendar;
```
rxaviers: Analogous to above.

rxaviers:
Great job so far!

Coding style... Your code uses arbitrary style. I'm not concerned with this right now. But, we'll have to fix all that at some point. Indentation, spaces, etc should follow http://contribute.jquery.org/style-guide/js/.

`cldr.attributes`... Please, do not change `cldr.attributes` ([code](https://github.com/jquery/globalize/pull/447/files#diff-7eb52b366866677666470e019283c8eaR46)). Instead, let's make something analogous to the numberingSystem (for numbers). In practice, this means instead of using 
`cldr.main("dates/calendars/{calendar}/dateTimeFormats")`, use `cldr.main([ "dates/calendars", calendar, "dateTimeFormats")`, where `calendar = fn( cldr )` (similar to this [code](https://github.com/jquery/globalize/blob/master/src/number/numbering-system-digits-map.js#L15)).

src/date/last-day-of-month.js
```diff
@@ -7,8 +7,13 @@ define(function() {
...
  */
-return function( date ) {
-	return new Date( date.getFullYear(), date.getMonth() + 1, 0).getDate();
+return function( gdate ) {
+  // no choice but to move forward one at a time
+  for (var tomorrow = gdate.nextDate(1); tomorrow.getMonth() === gdate.getMonth();){
+    gdate = tomorrow;
+		tomorrow = gdate.nextDate(1);
+  }
+  return gdate.getDate();
```
rxaviers: Can we move this manipulation into GDate?

src/date/parse.js
```diff
@@ -1,13 +1,11 @@
...
-	"./last-day-of-month",
+	"../core",
 	"./pattern-re",
 	"./start-of",
 	"../common/create-error/unsupported-feature",
-	"../util/date/set-month",
 	"../util/out-of-range"
-], function( dateIsLeapYear, dateLastDayOfMonth, datePatternRe, dateStartOf,
-	createErrorUnsupportedFeature, dateSetMonth, outOfRange ) {
+], function( Globalize, datePatternRe, dateStartOf,
```
rxaviers: We shouldn't need `Globalize` in here. Having calendars object in a different place (as commented above) will solve this.

src/date/parse.js
```diff
@@ -29,10 +28,33 @@ return function( value, tokens, properties ) {
...
+	ret = {
+		era: gdate.getEra(),
+		year: gdate.getYear(),
+		month: undefined,
+		date: undefined,
+		hour: date.getHours(),
+		minute: date.getMinutes(),
+		second: date.getSeconds(),
+		milliseconds: 0
+	};
```
rxaviers: Can we keep each field in its own local variable (not a ret property), like the example below. Because, we're not taking any benefit from having it inside `ret`. In fact, `ret.` becomes repetitive below.

```js
era = gdate.getEra();
year = ...
...
```

rxaviers: Plus, local variables are better minified. For example, `milliseconds` becomes an arbitrary variable like `a`. In the other hand, `ret.milliseconds` can't be changed. It becomes something like `a.milliseconds` when minified.

src/date/parse.js
```diff
@@ -236,24 +251,43 @@ return function( value, tokens, properties ) {
...
+		// if the date was invalid and gdate corrected it (like turning 31 Sep into 30 Sep),
+		// we want to reject it.
+		return null;
 	}
 
+  date = gdate.toDate();
+  date.setHours(ret.hour);
+  date.setMinutes(ret.minute);
+  date.setSeconds(ret.second);
+  date.setMilliseconds(ret.milliseconds);
```
rxaviers: Can we reverse this logic? I mean keep handling time fields the way we were before and then setting the date fields of `date` by using `setFns` functions with `gdate` values.

src/date/parse.js
```diff
@@ -236,24 +251,43 @@ return function( value, tokens, properties ) {
...
+	}
+	if ( ret.month == null ) {
+		ret.month = gdate.getMonth();
+	}
+	if ( ret.date == null && truncateAt.indexOf( MONTH ) !== -1) {
+		// month was defined but date was not
+		ret.date = 1; // gdate assumes 1 is the first date of the month
+	}
+	if ( ret.date == null ) {
+		ret.date = gdate.getDate();
```
rxaviers: Please, what changed compared to how we handled this previously?

src/calendars/calendar-for-locale.js
```diff
@@ -0,0 +1,57 @@
...
+	// the first one on both lists.
+	// But for now, just return the most preferred
+	if (cal) {
+		return cal.split( " " )[0];
+	}
+	// Return the default calendar
+	return "gregorian";
+};
+
+});
```
rxaviers: The algorithm looks great for now, except for one thing: I don't plan to require calendarPreference supplemental data. If such supplemental data is loaded, great. Use it. Otherwise, simply pick gregorian. Therefore, we should include `supplemental/calendarPreferenceData` path as another skip-validation-paths [here](https://github.com/dwachss/globalize/blob/320-nongregorian/src/date.js#L28).

rxaviers:
I left inline comments.

Variable names... There are a lot of abbreviated names, e.g., `cal`, `d`, `m` (especially in the gdate code). All variable names should use full words (names should be descriptive but not excessively so).

Third party code... If there is any code borrowed from different libraries, we must mention them. So, please let me know.

rxaviers:
Thanks so far and just let me know on any questions.

dwachss:
Thanks for all the comments and code review. I'm out of town for the next week (new granddaughter!) but I will go through it all when I get back.

rxaviers:
Congratulations and thanks for letting me know.

dwachss:
I think I've incorporated everything into the newest commit.
I'm happy to make stylistic changes to fit jQuery's needs; I figure that if it passed jscs it must be OK.

I changed the `".../{calendar}/..."` CLDR traversing to use `[..., calendar, ...]` without using `cldr.attributes`, but I disagree on this. You created an elegant mechanism for incorporating information in cldr paths; why not use it?

I'm still not happy with the way Gdate is defined. It ends up internal to dist/globalize/date.js, so it can't be used outside (say with a datepicker widget), and the calendar constructors have to be included in parse.js, rather than somehow only loading the ones the user needs. I'm not sure waht to do about that. I don't have any experience with AMD or bower to know how to make it its own module. Ideas would be appreciated.

rxaviers:
For the record, reviewing your changes is on my TODO list. I may be able to give you a position by mid next week.

rxaviers:
Just commenting to let you know I will need some more time to review.

dwachss:
No rush. I appreciate the time you're taking.

dwachss:
I added unit tests for the Hebrew and Islamic calendars. I added the code to support the Chinese calendar with its multiple intercalated leap months, but I don't know enough about it to actually implement it so it can be tested. 
I will work on that.

rxaviers:
Quick comment about testing... [ICU online page](http://demo.icu-project.org/icu4jweb/flexTest.jsp) could be used for checks/comparison.

dwachss:
Thank you for that link. 
What is the relationship between ICU and CLDR? They are not the same organization as far as I can tell; the CLDR documentation references what ICU does.

rxaviers:
ICU is an implementation (for C, C++ and Java) that uses CLDR (and that has actually originated CLDR).

ashensis:
IBM was commiting to contribute Islamic/Hebrew calendar/date picker to JQuery.
It looks that a few steps had been done already.
I am interested whether these Hebrew-date and Islamic-date classes written by dwachss were intended to from start to be coupled with JQuery date picker, do they posess all auxiliary functions to make support for corresponing national date pickers.
If the answer to aforementioned question is positive, could we contribute the corresponding date picker support to JQuery?
Again, why these 2 calendars didn't find their way in Globalization 1.0 yet, what are the plans regarding these 2 calendar classes.

tomerm:
First of all a couple of words on the expected functionality: 
1. From classes implementing calendaring logic we expect more or less standard API which should serve following purposes:
    a. Conversion of date represented as a string from one calendaring system to enother
    b. Support integration of calendaring logic into JQuery date picker - https://jqueryui.com/datepicker/

2. From date picker (https://jqueryui.com/datepicker/) side we expect ability to show the date inside date picker (both input field and rectangular area on which usually one month of the calendar is graphically displayed) based on different calendaring systems (type of calendar [Thai, Islamic, Hebrew etc.] can be probably passed as a parameter to date picker widget during creation / initialization.

Second, our intention is to understand the stage of current discussion. From https://github.com/jquery/globalize/issues/320 it is clear that there is an intention to support non Gregorian calendars. It is less clear why this specific pull request has not been integrated yet into Globalize JQuery package.  Are there any reservations related to submitted code ? What is missing ? 

Finally, already now it is clear that at the very least following items are missing in the current submission:
a. Integration of non Gregorian calendar logic into JQUery date picker
b. Support for Hebrew numerals in Hebrew calendar (https://en.wikipedia.org/wiki/Hebrew_numerals)
c. Additional types of non Gregorian calendars 
We are willing to contribute those items and want to better understand if we should base our work on top of @dwachss contribution or develop everything from 0.

rxaviers:
Hello @ashensis and @tomerm.

First of all thanks for your wiliness to contribute, which is very welcome, and sorry for my delayed answer. I suggest your contribution continues @dwachss work instead of beginning something from 0.

Some preliminary points for clarification:
- The Globalize scope (in this particular subject) is about formatting dates from JavaScript (i.e., instance of `Date()`) into its corresponding localized String on any of the calendars we support, i.e., gregorian and hopefully soon non-gregorian calendars too. It's also about the reverse operation, i.e., parsing the localized String into JavaScript `Date()` instance.
- The datepicker, i.e., the UI widget that renders the calendar and that manages user input is not part of Globalize. It's part of the [jQuery UI](https://github.com/jquery/jquery-ui/) and @scottgonzalez or @jzaefferer can guide you out here.

About this PR... All logic for calendrical calculations should be made by an independent library (currently living in this repository, but planned to live elsewhere), which we called GDate and its API is drafted [here](https://gist.github.com/dwachss/4f9a6c77c8feb8e2ad09). Globalize is going to use this library (or any other library that provides the same API) for converting calendars when formatting or parsing dates.

> I do not see methods like 'firstDayOfMonth', ' 'firstDayOfWeek' etc.

All calendrical calculations should be made by an independent library called GDate (see paragraph above).

Note though you can use Globalize for either to format or to parse these fields. See http://www.unicode.org/reports/tr35/tr35-dates.html#Date_Field_Symbol_Table for the appropriate pattern that you could use for these. Another related doc https://github.com/jquery/globalize/blob/master/doc/api/date/date-formatter.md.

> 1. From classes implementing calendaring logic we expect more or less standard API which should serve following purposes:
> a. Conversion of date represented as a string from one calendaring system to enother

The calendrical calculations library scope (GDate) is defined in the link mentioned above.  It's not in the scope of GDate to parse Strings though. It was planned to be kept as simple as possible. It's core is calendrical calculations. The scope of parsing strings is of Globalize.

By the way, what's your reference for the mentioned "standard API" please?

> It is less clear why this specific pull request has not been integrated yet into Globalize JQuery package. Are there any reservations related to submitted code ? What is missing ? 

Excellent question. This PR is already working to format and to parse some non-greg calendars. But, it isn't complete. The next step is to finish elaborating the GDate API and have Globalize to transparently use either a lightweight embedded GDate for gregorian calendars or the external GDate for non-greg calendars.

> b. Support for Hebrew numerals in Hebrew calendar (https://en.wikipedia.org/wiki/Hebrew_numerals)

Hebrew numbers are algorithmic, which isn't supported by Globalize. Globalize supports numeric digits only, e.g., latin, arabic, thai, bali, etc.

Would you like to contribute Globalize support for algorithmic numbers?

> c. Additional types of non Gregorian calendars 

Additional non greg calendars are welcome.

Did I answer to your questions? Please, just let me know if there's anything I can still answer. I hope Danny (@dwachss) can also weigh in.

PS: I'm skipping to answer the questions related to UI widget datepicker.

tomerm:
Hello Rafael,

   Thanks a lot for your reply.  We obviously don't want to start from 0 
if there are some ongoing efforts. This is one the primary reason for all 
the latest inquires discussed in this thread.

 >> Would you like to contribute Globalize support for algorithmic numbers 
(Hebrew numerals) ?
Yes, this is our intention.  The logic is relatively simple and will be 
submitted for review in Q4 2015

>> Additional non greg calendars are welcome.
@Mohamed & @Ramy please share your plans to add additional non Gregorian 
calendars commonly used in Arabic speaking countries to Jquery. Thanks. 

Best Regards,

Tomer Mahlin
Bidi Champion, Globalization Manager, GCoC
Bidi Development Lab


Phone: +972-2-6491784 | Mobile: +972-54-3368122 E-mail: tomerm@il.ibm.com

IBM R&D Labs
Malcha Technology Park
Jerusalem 96951
 Israel

rxaviers:
Excellent @tomerm, @Mohamed, and @Ramy. I'm looking forward to it. Please, just let me know if there's any input you need from me.

ashensis:
Hi Rafael.
Please look at new feature regarding  Globalize support for algorithmic numbers that I got created:
https://github.com/jquery/globalize/issues/537

dwachss:
I've pushed a new commit that changes the interface for GDate to separate month (which is now a number, as rxaviers has been arguing for) and monthType, which is a string that will be undefined for most calendars, and "leap" for lunisolar leap months. 
It became clear that this would be necessary when I tried to implement the Chinese calendar, The "monthType" parameter uses the terms from the CLDR monthPattern element ( http://www.unicode.org/reports/tr35/tr35-dates.html#monthPatterns_cyclicNameSets ).
See the updated GDate API gist at https://gist.github.com/dwachss/4f9a6c77c8feb8e2ad09

eximiusdevelopment:
Hi everyone,

I'm looking for support for the Persian calendar, is somebody already working on this by any chance?
Is there any chance that this branch (from dwachss) will be merged with master before the end of the year?

Regards,
Michael

rxaviers:
Hi @eximiusdevelopment, I don't think this is going to be ready and released by the end of the year, but this is an important feature that Globalize must support and all help is very much appreciated. Do you want to submit a PR extending this one with the support for Persian calendar?

Thanks.

eximiusdevelopment:
Hi Rafael,

That is ok, I have a branch but I'm new to this stuff so I have not yet created any tests for it. We have forked the branch of dwachss and added the Persian calendar.  But will this branch be merged into master in the future? Because I depend on the Gdate implementation.

We are currently working together with an Iranian company to get the dates validated, because we have taken the logic from existing implementations (keithwood I think and JDate) but we want to make sure it is solid. When that is done we will send a PR.


Regards,
Michael

dwachss:
That's great. The more calendars we can get implemented the better. Is this the Solar Hejri calendar (https://en.wikipedia.org/wiki/Solar_Hijri_calendar) ? I am working on getting the astronomical calculations that are needed for all these calendars and I'd like to see what you have. Looking forward to your PR.Danny 


    On Wednesday, December 9, 2015 11:01 AM, eximiusdevelopment <notifications@github.com> wrote:
 

 Hi Rafael,That is ok, I have a branch but I'm new to this stuff so I have not yet created any tests for it. We have forked the branch of dwachss and added the Persian calendar. But will this branch be merged into master in the future? Because I depend on the Gdate implementation.We are currently working together with an Iranian company to get the dates validated, because we have taken the logic from existing implementations (keithwood I think and JDate) but we want to make sure it is solid. When that is done we will send a PR.Regards,
Michael —

eximiusdevelopment:
Hi Danny,

First of all thank you for your initial code of the Gdate, without that code we would not be able to do this in a short time.
The calendar we are implementing is the Persian Jalali calendar (https://en.wikipedia.org/wiki/Jalali_calendar), which is used in Iran.

Regards,
Michael

src/date/tokenizer-properties.js
```diff
@@ -76,10 +84,35 @@ return function( pattern, cldr ) {
...
+					properties[key] = dateMonthNames (
+						properties[key],
+						chr,
+						length,
+						properties.calendar,
+						cldr
+					);
+				} else {
+
+				// record the possible expansion for numeric months
```
rxaviers: let's indent this according to the below.

src/date/tokenizer.js
```diff
@@ -132,20 +133,48 @@ return function( value, numberParser, properties ) {
 			}
 		}
 
+		function monthNumber() {
+			var match, monthPatterns, reSource, type;
+			if ( length <= 2 ) {
+
+			// Unicode equivalent to \d+
```
rxaviers: Let's indent this according to the below.

test/unit/date/format.js
```diff
@@ -61,7 +61,7 @@ QUnit.assert.dateFormat = function( date, pattern, cldr, expected ) {
 	var pad,
 		numberFormatters = [],
 		properties = formatProperties( pattern, cldr );
-
+
```
rxaviers: trailing space

src/date/format-properties.js
```diff
@@ -103,19 +109,25 @@ return function( pattern, cldr ) {
...
+				properties.months[ chr ][ length ] = dateMonthNames(
+					properties.months[ chr ][ length ],
+					chr,
+					length,
+					properties.calendar,
+					cldr
+				);
+				if ( length <= 2 ) {
 					formatNumber = true;
 				}
```
rxaviers: Can we keep it as `else` above and have the inner code afterwards?

rxaviers: Is `properties.months` now when `formatNumber = true`?

src/date/format.js
```diff
@@ -97,7 +96,12 @@ return function( date, numberFormatters, properties ) {
 			// Quarter
 			case "Q":
 			case "q":
-				ret = Math.ceil( ( date.getMonth() + 1 ) / 3 );
+
+        // There's no good way to do this with a generalized date.
```
rxaviers: indentation

src/date/format.js
```diff
@@ -97,7 +96,12 @@ return function( date, numberFormatters, properties ) {
...
+				// We have to approximate this, assuming a 12-month year and
+				// month numbers that correspond to the correct period of the year
+				ret = Math.min( parseInt( gdate.getMonth(), 10 ), 12 );
```
rxaviers: Please, what's the return value of `.getMonth()` is it a string? Where does this string comes from? I guess this comes from CLDR, the key of months object. Just confirming. Is this always a digit?

src/date/month-names.js
```diff
@@ -0,0 +1,82 @@
...
+ *
+ */
+return function( months, type, length, calendar, cldr ) {
+	var i, month, monthParts, monthPatterns, pattern;
+
+	if ( months ) {
+
+	// fastest (native) way to clone an object
+		months = JSON.parse( JSON.stringify( months ) );
+	}
```
rxaviers: I'm wondering if we could avoid this.

src/date/start-of.js
```diff
@@ -7,17 +9,21 @@ define(function() {
...
 	switch ( unit ) {
 		case "year":
-			date.setMonth( 0 );
-		/* falls through */
 		case "month":
-			date.setDate( 1 );
-		/* falls through */
 		case "day":
 			date.setHours( 0 );
 		/* falls through */
```
rxaviers: Do we still need the switch?

src/date/format.js
dwachss: getMonth returns a number, so the parseInt is a bug. I'll remove it.

src/date/tokenizer-properties.js
dwachss: I'm not sure what you mean.

src/date/tokenizer.js
dwachss: I'm not sure what you mean.

test/unit/date/format.js
dwachss: corrected

src/date/format-properties.js
dwachss: Unfortunately, the CLDR seems to specify that Chinese leap months get names like "6bis" or "06bis" even for "numeric" months. So properties.months is still needed. 
The code initially sets it from CLDR, then uses dateMonthNames to update the names of the leap months.

I can move that final `if` back into the else.

src/date/month-names.js
dwachss: are you talking about the clone?  `months` is returned from `cldr.main('dates/calendar')` and is the actual data from the global CLDR object. I'm hesitant to change that directly; I don't know if any other part of the program relies on the month names in their original form.
You know cldr.js; if it's OK to modify it, we can drop that line.

src/date/start-of.js
dwachss: I don't think we need the startOf function at all. I haven't eliminated it yet, but it should be.

Ashamandi:
Hello @dwachss and @rxaviers , Could you please tell me what is done at supporting Hijri calender so far ? because I am interesting in contributing in this support, and as I see you don`t add it yet , Am I right?

dwachss:
Hijiri is the traditional Japanese calendar, right? I haven't done that, only Hebrew, Islamic and Chinese. Look at the gdate documentation and feel free to contribute. 
The file astronomy.js has routines for new moons and solstices if you need that. 
Danny

Ashamandi:
Sorry but I mean by Hijri calendar the Islamic one which is a lunar calendar consisting of 12 months in a year of 354 or 355 days, Do you support it ??

Ashamandi:
I go to /gdate/Islamic-date.js 
I find it is about tabular calender
+// Islamic tabular calendar (http://en.wikipedia.org/wiki/Tabular_Islamic_calendar)

but I want to support muslium hijri calender 
https://en.wikipedia.org/wiki/Islamic_calendar

dwachss:
I see. I have not implemented the astronomic Islamic calendar, but your help would be appreciated. As I said, the astronomy.js file has routines for new moons, but the details of the algorithm and adjusting for Mecca local time needs to be done.

Ashamandi:
@dwachss, Sorry for this misunderstanding, I found that you had already implemented what I was looking for so I would like to thank you for your great work, 
and I have a question "When will this PR be ready and released so we can use this version of Globalize at Jquery-ui and create an islamic datepicker widget?" ... Thanks in advance

scottgonzalez:
This will be released as part of jQuery UI 1.13.0, but we never assign or publicly forecast release dates.

Ashamandi:
@scottgonzalez, thanks so much for your quick reply.

rxaviers:

Issue #405, opened 2015-02-25
Suggestion: Simpler loading of CLDR with loadFromPath() and fix leaking internal path information


hannes-landeholm:
I don't understand why people writing applications that use globalize.js have to bother with the internal path standard that CLDR uses and understand what module uses what paths? Unless I'm mistaken, in 99% of all cases users will just put the CLDR data somewhere and then use globalize.js without changing any paths.

* It's a headache for web developers as they have to read documentation about CLDR and the "getting started guide" instead of just plugging in globalize.js and the CLDR data and using it.
* It leaks information to a layer that shouldn't even ideally have access to the information. What CLDR data certain globalize.js modules use is not relevant for applications using globalize.js and they should not define them at all.
* It cements the CLDR path standard so if it ever changes in the future I have to rewrite my application with new paths instead of just updating globalize.js and my CLDR json mirror.

Instead of doing this insanity: 

    $.when(
      $.get( "/static/cldr/main/en/ca-gregorian.json" ),
      $.get( "/static/cldr/supplemental/likelySubtags.json" ),
      $.get( "/static/cldr/supplemental/timeData.json" ),
      $.get( "/static/cldr/supplemental/weekData.json" )
    ).then(function() {
    
      // Normalize $.get results, we only need the JSON, not the request statuses.
      return [].slice.apply( arguments, [ 0 ] ).map(function( result ) {
          return result[ 0 ];
      });
    
    }).then( Globalize.load ).then(function() {
    
      // Your code goes here.
    
    });

Why not just simply provide this interface which does all those things itself instead?

    Globalize.loadFromPath("/static/cldr/", ["en"], ["date"], function(success) {
        // Your code goes here.
    });

* More future proof.
* Much easier and simpler to use.

In addition, I think the README should be updated so this way is strongly recommended and any other method is hidden in some "advanced use" section. The table describing what modules uses what CLDR data should be removed or hidden.

Thank you.

rxaviers:
Hi, your feedback is very much appreciated.

[...]

PR #452, opened 2015-06-12
For easier usage added function to load cldr data from file

romanWienicke:
After a lot of time trying to get the system running i figured out that it would be good to get a method to load the cldr data from a given file.
I added the getCldrData function, what do you think about it?
(It's my first contribution so I'd be glad to get some information if everything is ok)

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,6 +264,18 @@ to you in different flavors):
 
   [Read more...](doc/api/core/load.md)
 
+- **`Globalize.getCldrData( [baseUrl], [fileName] )`**
+  
+  This method loads the cldr data with JQuery from a given JSQL local file synchroniously.
+  
+  Example:
+  the base setup:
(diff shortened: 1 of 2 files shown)


rxaviers:
Hi @romanWienicke, thank you for your pro activity by submitting this PR.

The big problem of trying to handle loading is trying to support all the various different environments. For more information, see https://github.com/jquery/globalize/issues/405#issuecomment-75952649.

Having said that, I am open to all ideas that help easing the content loading ramp up. For example, #345, [cldr-data AMD plugin](https://github.com/rxaviers/cldr-data-bower/tree/amd-plugin).

How can we get to the level of abstraction you are proposing, but without overloading Globalize? As a plug-in library?

rxaviers:
Closing as duplicate of #405. But, still open for your thoughts.

PR #456, opened 2015-06-24
Avoid re-using a variable for a different purpose

edg2s:
fraction starts off as an object then becomes a string

### DIFF
--- a/src/currency/supplemental-override.js
+++ b/src/currency/supplemental-override.js
@@ -9,15 +9,14 @@ define([
  */
 return function( currency, pattern, cldr ) {
 	var digits,
-		fraction = cldr.supplemental([ "currencyData/fractions", currency ]) ||
+		fraction = "",
+		fractionData = cldr.supplemental([ "currencyData/fractions", currency ]) ||
 			cldr.supplemental( "currencyData/fractions/DEFAULT" );
 
-	digits = +fraction._digits;
+	digits = +fractionData._digits;
 
 	if ( digits ) {
-		fraction = "." + stringPad( "0", digits ).slice( 0, -1 ) + fraction._rounding;
-	} else {
-		fraction = "";
+		fraction = "." + stringPad( "0", digits ).slice( 0, -1 ) + fractionData._rounding;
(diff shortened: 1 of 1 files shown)


edg2s:
I signed the CLA, need to retrigger?

src/currency/supplemental-override.js
```diff
@@ -8,16 +8,13 @@ define([
...
 return function( currency, pattern, cldr ) {
-	var digits,
-		fraction = cldr.supplemental([ "currencyData/fractions", currency ]) ||
-			cldr.supplemental( "currencyData/fractions/DEFAULT" );
-
-	digits = +fraction._digits;
+	var fraction = "",
+		fractionData = cldr.supplemental([ "currencyData/fractions", currency ]) ||
+			cldr.supplemental( "currencyData/fractions/DEFAULT" ),
+		digits = +fractionData._digits;
```
rxaviers: `digits` depends on `fractionData` to be set. Therefore, I prefer it on a new line. Can you revert that part, please? (this is, `var digits,` then `digits = `)

rxaviers:
Thanks for your contribution. PS: The CLA checker will only retrigger on new commits, but I can check it manually, don't worry.

src/currency/supplemental-override.js
edg2s: sure

rxaviers:
I'm cc'ing @jquery/globalize to review this code to avoid me being biased to think the former is cleaner given the fact I wrote it. :)

jzaefferer:
Change looks good to me. Reusing a variable for different types seems bad indeed.

As for the CLA, it looks like there's multiple signatures with the same email address, which our automatic check [doesn't handle properly](https://github.com/jquery/jquery-license/issues/32). Changing the git author to the full name should fix the issue, since that's what the first signature had. Either way we've got a valid signature.

rxaviers:
Thanks @edg2s for your contribution. Merged.

rxaviers:
@jzaefferer thanks for your review.

PR #462, opened 2015-07-15
Dynamically augment availableFormats

### LINKED ISSUE #271: Date (skeletons): dynamically augment availableFormats

vectart:
fixes #271

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,3 +7,4 @@ before_install:
 install:
 - npm install
 - bower install
+sudo: false
--- a/src/date/expand-pattern.js
+++ b/src/date/expand-pattern.js
@@ -1,7 +1,8 @@
 define([
 	"../common/format-message",
-	"../common/create-error/invalid-parameter-value"
-], function( formatMessage, createErrorInvalidParameterValue ) {
+	"../common/create-error/invalid-parameter-value",
+	"./pattern-re"
+], function( formatMessage, createErrorInvalidParameterValue, datePatternRe ) {
 
 /**
  * expandPattern( options, cldr )
@@ -26,7 +27,7 @@ define([
  */
 
 return function( options, cldr ) {
-	var dateSkeleton, result, skeleton, timeSkeleton, type;
+	var dateSkeleton, result, skeleton, timeSkeleton, type, pattern;
 
 	function combineDateTime( type, datePattern, timePattern ) {
 		return formatMessage(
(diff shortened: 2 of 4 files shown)


rxaviers:
Awesome! All I see for now are some style issues, which I'm going to comment inline.

.travis.yml
```diff
@@ -7,3 +7,4 @@ before_install:
 install:
 - npm install
 - bower install
+sudo: false
```
rxaviers: cc @jzaefferer @scottgonzalez to comment on that

src/date/expand-pattern.js
```diff
@@ -38,6 +39,89 @@ return function( options, cldr ) {
 		);
 	}
 
+	function getBestMatchPattern( path, skeleton ) {
+		var availableFormats, ratedFormats, format;
+
+		pattern = cldr.main([ path, skeleton ]);
+
+		if (skeleton && pattern === undefined) {
```
rxaviers: Spaces `if ( skeleton && pattern === undefined ) {`

rxaviers: If assume it's safe to use `!pattern` in here.

src/date/expand-pattern.js
```diff
@@ -38,6 +39,89 @@ return function( options, cldr ) {
 		);
 	}
 
+	function getBestMatchPattern( path, skeleton ) {
+		var availableFormats, ratedFormats, format;
```
rxaviers: I assume you want to define `pattern` as a variable local to this function, so it's not mixed up with `pattern` defined by the parent function.

src/date/expand-pattern.js
```diff
@@ -38,6 +39,89 @@ return function( options, cldr ) {
...
+
+		pattern = cldr.main([ path, skeleton ]);
+
+		if (skeleton && pattern === undefined) {
+			availableFormats = cldr.main([ path ]);
+			ratedFormats = [];
+
+			for (format in availableFormats) {
```
rxaviers: Spaces `for ( format in availableFormats ) {`

rxaviers: There are more occurrences like this below...

src/date/expand-pattern.js
```diff
@@ -38,6 +39,89 @@ return function( options, cldr ) {
...
+
+		if (a.length === b.length) {
+			distance *= 0.5;
+		}
+
+		return distance;
+	}
+
+	function augmentFormat( requestedSkeleton, bestMatchFormat ) {
+		var originalBestMatchFormat, i, t, j, k, l;
```
rxaviers: variable names: http://contribute.jquery.org/style-guide/js/#naming-conventions

src/date/expand-pattern.js
```diff
@@ -38,6 +39,89 @@ return function( options, cldr ) {
...
+
+		originalBestMatchFormat = bestMatchFormat;
+		requestedSkeleton = requestedSkeleton.match(datePatternRe);
+		bestMatchFormat = bestMatchFormat.match(datePatternRe);
+
+		for (i = 0, l = bestMatchFormat.length; i < l; i++) {
```
rxaviers: @jquery/globalize what's your thought about it? Is it worth for performance reasons or it's safe to compare to length on each iteration?

rxaviers: @vectart yeap, it's safe to compare to length directly `for ( i = 0; i < bestMatchFormat.length; i++ ) {`, length won't iterate through the array to calculate itself each time. So, you can get rid of 2 variables.

rxaviers:
@vectart I left a couple of comments, please just let me know on any questions and thanks so far.

vectart:
@rxaviers Thanks for the useful comments, I've updated PR according to them

.travis.yml
jzaefferer: As long as its in a separate commit. Should also enable caching, see, for example, https://github.com/jquery/qunit/pull/835

rxaviers: @vectart please could you include this? :point_up:

rxaviers:
Please, could you rebase on master (`git rebase master`)? I've updated jscs and it will help us to find the coding style issues.

src/date/expand-pattern.js
```diff
@@ -38,6 +39,95 @@ return function( options, cldr ) {
...
+		maxLength = Math.max( formatA.length, formatB.length );
+		minLength = Math.min( formatA.length, formatB.length );
+		distance = maxLength / minLength;
+
+		formatA = formatA.match( datePatternRe );
+		formatB = formatB.match( datePatternRe );
+		maxLength = Math.max( formatA.length, formatB.length );
+		minLength = Math.min( formatA.length, formatB.length );
+
+		for ( index = 0; index < minLength; index++ ) {
```
rxaviers: It's ok using `i` instead of `index` here. :)

src/date/expand-pattern.js
```diff
@@ -38,6 +39,95 @@ return function( options, cldr ) {
...
+					rate: compareFormats( skeleton, format )
+				});
+			}
+
+			ratedFormats.sort( function( formatA, formatB ) {
+				return formatA.rate - formatB.rate;
+			});
+
+			if ( ratedFormats.length ) {
+				pattern = augmentFormat( skeleton, ratedFormats[0].pattern );
```
rxaviers: space `ratedFormats[ 0 ]`

rxaviers:
Interestingly, the augmentation for `Globalize("en").formatDate(new Date(), {skeleton: "hhmm"})` worked just fine: `'12:34 AM'`. Although, it didn't work for `Globalize("en").formatDate(new Date(), {skeleton: "HHmm"})`: `'0033'` (`'00:33''` expected). Please, could you check and also add this as a new test case?

rxaviers:
Another worth doing test is checking against ICU... One difference I found is: `EHmss` gives 'Wed, 20:38:05' (seems to have chosen `E` alone and `ms` alone, then augmented `ms` into `mss`, then glued them) opposed to `Wed 20:38:05` (seems to have chosen `EHms`, then augmented it into `EHmss`) on ICU (http://demo.icu-project.org/icu4jweb/flexTest.jsp?pat=EHmss&_=en_US)

src/date/expand-pattern.js
```diff
@@ -57,16 +147,20 @@ return function( options, cldr ) {
...
-						timeSkeleton
-					])
+				dateSkeleton = getBestMatchPattern(
+					"dates/calendars/gregorian/dateTimeFormats/availableFormats",
+					dateSkeleton
 				);
+				timeSkeleton = getBestMatchPattern(
+					"dates/calendars/gregorian/dateTimeFormats/availableFormats",
+					timeSkeleton
+				);
```
rxaviers: You might need use `getBestMatchPattern()` for each date and time before the `type` assignment above. For example, `MMMMh` should give `July, 8 PM` (type medium) instead of `July at 8 PM` (type long). Because, "MMMM" will be converted into "LLLL".

rxaviers:
@vectart just let me know when you have progress, so we can land this fix. Thanks for your great work so far.

Strate:
How about to use builtin `new Intl.DateTimeFormatter().resolved.pattern` to get best-fit pattern?

PR #481, opened 2015-08-14
Implement app example using webpack

### LINKED ISSUE #464: Full stack example: webpack + Globalize
PR #481

Provide better documentation by creating a full stack example using webpack and Globalize.

Ref #441 #467

##### Development
- **Easy development**, Globalize is a powerful and flexible library, but it can be frustrating to set it up (e.g., loading required CLDR data on development, using compiled data instead on production). It should be easy for developers to build apps using Globalize. Ideally, they should be able to use Globalize (declaratively) without worrying about any of these setups.
- **Fast Live Reload**, Webpack allows writing code in source code formats that browsers don't natively interpret (e.g., coffee script, ES6, less or sass styles, etc) and have them to display it in real time by pre-processing the code into a compatible JavaScript and by having browsers to auto refresh on a process called live reload (HMR Hot Module Replacement), which is a very powerful tool. But, it's frustrating when the pre-processing takes too much time and doesn't get responsive. Ideally, Globalize on development shouldn't weight the process, it should allow for very fast updates from its part.

##### Production

- **Optimized for Production**, use precompiled formatters/parsers: #398
  - Small. Avoid including unnecessary i18n data. For example, doesn't include unnecessary translation messages, doesn't include unnecessary functionality data (e.g., doesn't include calendar information if not formatting dates, doesn't include currency data if not formatting currencies, etc), doesn't include unnecessary data within the same functionality (e.g., doesn't include month names if formatting dates using numbered months only). Thefore, no bandwidth is wasted.
  - Fast. Have all formatters (numbers, currencies, datetimes, relative time) generated/preprocessed at built time. This is, traversing CLDR data and generating the formatters will happen during build time and will be precompiled for runtime. Therefore, no CPU clocks are wasted on the client.
- **Compiled-Globalize-Code in separate chunks**, allow grouping compiled-globalize code in separate chunks.

rxaviers:
WIP fix for #464

Requires:
- [x] #398
- [x] [rxaviers/globalize-webpack-plugin](https://github.com/rxaviers/globalize-webpack-plugin)

ToDo:
- [x] Declarative development - don't worry about loading CLDR or using pre-compiled instead.
- [x] Fast Live Reload - use dynamically-created-generators. (via `npm run start` in this example)
- [x] Optimized for Production (it's using a temporary branch with runtime support - need update when #398 lands).
- [x] Compiled-Globalize-Code in separate chunks. (via `npm run build` in this example)
  - [x] for the development locale (`en` in this example).
  - [x] for each supported locale (`en`, `es` and `zh` in this example).
- [x] Example instructions README.md.
- [x] Using the same hello world example for the app. While the example must be concise, ideally we should provide one closer to real life. Should we? Ideas?
  - In the end, the demo implemented here was not very much a real life thing. But, at least, slightly better than the hello world one.
- [x] Proper translation messages for supported locales.
- [x] Update README linking to this example.
#### About this implementation:
- Application doesn't require developer to handle CLDR loading as can be observed at https://github.com/jquery/globalize/blob/fix-464-webpack-example/examples/full-app-npm-webpack/app/index.js. Globalize plugin will handle that.
- Webpack relevant configuration is
  https://github.com/jquery/globalize/blob/fix-464-webpack-example/examples/app-npm-webpack/webpack-config.js#L38-L44

``` js
new globalizePlugin({
    production: options.production, // true: production, false: development
    developmentLocale: "en", // locale to be used for development.
    supportedLocales: [ "en", "es", "zh" ], // locales that should be built support for.
    messages: "messages/[locale].json", // messages (optional)
    output: "globalize-compiled-[locale].[hash].js" // build output.
})
```

_production_ is a boolean that tells the plugin whether it's on production mode (i.e., to build the precompiled globalize data) or not (i.e., to be in development mode - will use Live AutoReload HRM).

_developmentLocale_ tells the plugin which locale to automatically load CLDR for and have it set as default locale for Globalize (i.e., `Globalize.locale(developmentLocale)`).

_supportedLocales_ tells the plugin which locales to build/produce compiled-globalize-data for.

_output_ is the name scheme of the built files.
##### Development mode

Start the Live AutoReload server by running the command below.

```
npm run start
```

Then, point your browser to `http://localhost:8080/`.
##### Production mode

Build the production bundles by running the command below.

```
npm run build
```

The build files are placed in `dist`. Point your browser at `./dist/index.html` to check.


sompylasar:
+ Using a webpack loader to load the messages (not just from plain JSON).

The real-world use case. I'm migrating a large code base to a set of modern tools, including webpack and, probably, Globalize for i18n, but my messages are in PHP (and that cannot be changed during this migration because they are reused by the backend i18n system). We're currently developing a loader that provides the messages to our frontend i18n system in the required format, and it would fit well if we decide to replace our frontend i18n with Globalize.

+ Runtime language switching (and asynchronous loading of the required locale which is not loaded during initial page load).

The real-world use case is obvious: language switching without page reload.

rxaviers:
> - Using a webpack loader or a function to load the messages (not just from plain JSON).

I see. Can you provide examples of the not-just-from-plain-JSON-alternatives please?

> Runtime language switching (and asynchronous loading of the required locale which is not loaded during initial page load).

:thumbsup:

sompylasar:
@rxaviers I'll give you an example but please do not write a specific loader for my specific example. A generic ability to specify a loader (or maybe better a plain function which could delegate to any other code if needed) would be better.

Our i18n system is Yii-based and uses a notion of per-module "category" and "key" (not just "key" which Globalize uses AFAIK), so a transform is required. But the system is highly customized to support multi-level overrides and fallbacks for each language, so this should be somehow handled by our custom loader (we actually delegate to the PHP component which knows how to do overrides and fallbacks, it does this for the server-side).

Here's an example of a single message file (of a single "category" for a "dialogs" module):
```php
<?php

return [

	"_meta" => [
		"selector" => [
			"en",
		],
	],

	"dlg.message.title" => "Message",

	"dlg.message.button.close" => "Close",

	"dlg.error.title" => "Error",

	"dlg.info.title" => "Info",

	"dlg.confirm.title" => "Confirmation",

	"dlg.confirm.button.yes" => "Yes",

	"dlg.confirm.button.no" => "No",

	"dlg.prompt.title" => "Input",

	"dlg.prompt.button.yes" => "OK",

	"dlg.prompt.button.no" => "Cancel",

];

```

It's time for one more unexpressed thought that I've written about before. We need to be able to load different sets of messages for different apps based on a single platform (each app has one or more webpack entry points). Imagine two different applications with a set of common modules, each module depends on its own "category" with translations, each translation may be overridden at the application level. We should be able to build these apps separately and bundle only the translations being required by each app's components.

rxaviers:
Sure, the messages option can be made to either accept the string (as documented above) or a function that passes `locale` as its first argument. For example:

```js
new globalizePlugin({
    production: options.production, // true: production, false: development
    developmentLocale: "en", // locale to be used for development.
    supportedLocales: [ "en", "es", "zh" ], // locales that should be built support for.
    messages: function( locale ) {
        // Messages in the JSON format for requested locale.
        return getMessagesFor( locale );
    }
    output: "globalize-compiled-[locale].[hash].js" // build output.
});
```

Will include that...

> We should be able to build these apps separately and bundle only the translations being required by each app's components.

The current implementation of [rxaviers/globalize-webpack-plugin](https://github.com/rxaviers/globalize-webpack-plugin) is smart to bundle the translations being required only.

sompylasar:
> The current implementation (rxaviers/globalize-webpack-plugin) is smart to bundle the translations being required only.

Wow, that's pretty cool if I get you correct! How would you require specific message files from component files then, not having a gigantic "messages" object?

sompylasar:
> the messages option can be made to either accept the string (as documented above) or a function that passes locale as its first argument.

Does it by chance support asynchronous loading e.g. via a Promise returned from the "messages" function?
How will that function understand what translation files have been required by the components the entry points of the current compilation depend on?

rxaviers:
Making myself clear, the current implementation requires a gigantic "messages" object to be available for the webpack plugin. But, the plugin is smart to use the messages being required only (and therefore, to generate the final bundle with the translations being required only).

> Does it by chance support asynchronous loading e.g. via a Promise returned from the "messages" function?

Async might be tricky in webpack. Most of the compilation plugins (for example, [[1](https://github.com/rxaviers/globalize-webpack-plugin/blob/b0.0.1/ProductionModePlugin.js#L117-L119)], [[2](https://github.com/rxaviers/globalize-webpack-plugin/blob/b0.0.1/ProductionModePlugin.js#L125-L147)]) are sync calls.

I'd like to understand how you handle your custom messages and where async comes in. For example, are you handling files locally or remote? Does the processing require async? We could chat on IRC (http://irc.jquery.org/, #globalize).

> How will that function understand what translation files have been required by the components the entry points of the current compilation depend on?

It's possible to pass, as another argument, the filename of the JS file being handled (in webpack terminology, the `request` variable):

```js
    // For example:
    messages: function( locale, request ) {
        // Messages in the JSON format for requested locale.
        // locale: E.g. "en", "en-GB", etc.
        // request:  E.g., "my-app/index.js", "my-app/components/foo.js", so your function
        //           could look into "my-app/index.messages.json" or
        //           "my-app/components/foo.messages.json" for example
        return getMessagesFor( locale, request );
    }
```

Did I answer your question?

sompylasar:
> Async might be tricky in webpack. Most of the compilation plugins (for example, [1], [2]) are sync calls.
> 
> I'd like to understand how you handle your custom messages and where async comes in. For example, are you handling files locally or remote? Does the processing require async? We could chat on IRC (http://irc.jquery.org/, #globalize).

The current implementation executes an external process (a PHP console command which composes the messages using our custom override logic and emits them to stdout) which is an async operation.

I'm sorry for confusing you, we are currently using a plugin, not a loader (we tried loader before but failed). Because of lack of documentation on webpack plugin features we are trying to learn from existing plugins (including yours from now). The current implementation looks hacky and requires us to include the i18n bundle separately, not using webpack autoloading system. The code is not public, but I can show you some generic pieces we came up with.

We use the `normal-module-factory` to gather the categories from `require` requests (each module requires a special module path, e.g. `i18n/my-category`, which can be captured by a RegExp). We have to replace the request with a dummy file.
```js
compiler.plugin('normal-module-factory', function (nmf) {
		langCategories = [];

		nmf.plugin('before-resolve', function (data, beforeResolveCallback) {
// ...
                                langCategories.push(langCategory);
                                data.request = __dirname + '/.i18n-dummy';
```

Then we use the `emit` compile step with asynchronous resolution (`compilationCallback`). We emit a compilation asset with a JS code that uses a globally available object to inject the translations it contains.
```js
compiler.plugin('emit', function (compilation, compilationCallback) {
// ...
            compilation.assets['some-prefix-' + language + '.js'] =  source;
//
```

> It's possible to pass, as another argument, the filename of the JS file being handled (in webpack terminology, the request variable):

Yes, this would be better, I think. This makes the "messages" function return a per-request object instead of the gigantic one, and Globalize then merges all the messages together, right?

rxaviers:
Webpack is very powerful, but it has been really laborious to develop the plugin. I've also been using another plugins as a baseline given the lack of doc, plus I have also been reading the source code and have counted with help from bebraw and sokra.

Thanks for sharing your excerpts. I understood that your `require` categories return a no-op source. So, I didn't understand how your `'some-prefix-' + language + '.js'` assets get loaded by your application? In other words, I didn't understand the link between the code that requires a category and the asset that is generated based on it.

> Yes, this would be better, I think. This makes the "messages" function return a per-request object instead of the gigantic one, and Globalize then merges all the messages together, right?

Ok, will include it or suggest something else after thinking more about it.

sompylasar:
> I understood that your require categories return a no-op source. So, I didn't understand how your 'some-prefix-' + language + '.js' assets get loaded by your application? 

Yes. That's the hacky flavor this solution gets. We access the translations via a singleton component which is populated with the messages. We have to side-load these files by a plain script tag or our own script loader, not by the webpack auto-loader. The messages get injected via a global variable shared between the core JS and the messages JS.

It looks like this (the code is pseudo-code):
```js
// some-prefix-en.js
window.globalI18nManager.addMessages("some-category", {
    "some-key": "Some message"
}, "en");

// bundle.js
window.globalI18nManager.setLanguage("en");
$someElement.text( window.globalI18nManager.formatMessage("some-category", "some-key") );
```

examples/app-npm-webpack/index-template.html
```diff
@@ -0,0 +1,80 @@
...
+    it.
+    -->
+    {%      } %}
+    <!-- <script src="{%=o.htmlWebpackPlugin.files.chunks[chunk].entry %}"></script> -->
+    {% } else { %}
+    <script src="{%=o.htmlWebpackPlugin.files.chunks[chunk].entry %}"></script>
+    {%
+        }
+    }
+    %}
```
rxaviers: It generates stuff like:

```html
    <script src="vendor.07c80e6dfcd89796c7d5.js"></script>
    
    <!--
    Load support for the `en` (English) locale.
    
    For displaying the application in a different locale, replace `en` with
    whatever other desired supported locale, e.g., `pt` (Portuguese).

    For supporting additional locales simultaneously and then having your
    application to change display dynamically, load the multiple files here.
    Then, use `Globalize.locale( <locale> )` in your application to dynamically
    set it.
    -->
    
    <!-- <script src="i18n/zh.07c80e6dfcd89796c7d5.js"></script> -->
    
    <!-- <script src="i18n/pt.07c80e6dfcd89796c7d5.js"></script> -->
    
    <!-- <script src="i18n/es.07c80e6dfcd89796c7d5.js"></script> -->
    
    <script src="i18n/en.07c80e6dfcd89796c7d5.js"></script>
    
    <script src="app.07c80e6dfcd89796c7d5.js"></script>
```

examples/app-npm-webpack/messages/en.json
```diff
@@ -0,0 +1,17 @@
...
+		"message-1": "An example of a message using mixed numbers \"{number}\", currencies \"{currency}\", dates \"{date}\", and relative time \"{relativeTime}\".",
+		"message-2": [
+			"An example of a message with pluralization support:",
+			"{count, plural,",
+			"    one {You have one remaining task}",
+			"  other {You have # remaining tasks}",
+			"}."
+		]
+	}
+}
```
rxaviers: @jzaefferer would you like to add support for `de`? :)

rxaviers: @isaacdurazo would you like to add support for `es`? :)

rxaviers:
@sompylasar your usecase should be tracked by https://github.com/rxaviers/globalize-webpack-plugin/issues/1 and https://github.com/rxaviers/globalize-webpack-plugin/issues/2

rxaviers:
Content team, @kswedberg, @arthurvr, if anyone has any spare time, it would be great to count with your continued help reviewing this PR as well. Specially in:
- https://github.com/jquery/globalize/pull/481/files#diff-96890900e38239964de166018af8e295
- https://github.com/jquery/globalize/pull/481/files#r37784158

sompylasar:
@rxaviers Thanks!

kswedberg:
thanks, @rxaviers . I should be able to do a quick review this evening.

examples/app-npm-webpack/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "globalize-full-app-npm-webpack",
```
arthurvr: Dunno how we generally do that, but if a package is marked `private` giving a `name` is kinda moot.

examples/app-npm-webpack/index-template.html
```diff
@@ -0,0 +1,80 @@
+<!doctype html>
+<head>
+    <meta charset="utf-8">
```
arthurvr: This indentation seems wrong.

examples/app-npm-webpack/index-template.html
```diff
@@ -0,0 +1,80 @@
+<!doctype html>
+<head>
```
arthurvr: missing a `<html>` tag?

examples/app-npm-webpack/package.json
rxaviers: :+1:

rxaviers: ef8dd29bf0efe5d13f4b4d99d979436f9fddfc7a

arthurvr:
Skimmed the text really quickly and it looks pretty alright. Nice work @rxaviers!

examples/app-npm-webpack/index-template.html
rxaviers: a85966e

examples/app-npm-webpack/index-template.html
rxaviers: a85966e

rxaviers:
Thanks for your review so far @arthurvr. Ok, @kswedberg thanks in advance.

examples/app-npm-webpack/app/index.js
```diff
@@ -0,0 +1,53 @@
...
+// Messages.
+document.getElementById( "message-1" ).innerHTML = Globalize.formatMessage( "message-1", {
+	number: currencyFormatter( 69900 ),
+	currency: dateFormatter( new Date() ),
+	date: numberFormatter( 12345.6789 ),
+	relativeTime: relativeTimeFormatter( 0 )
+});
+
+document.getElementById( "message-2" ).innerHTML = Globalize.formatMessage( "message-2", {
+	count: 3
```
sompylasar: Maybe for the demo purposes also increment the `count` from `0` to some number and back to `0`? This would demonstrate how plurals work with different values.

examples/app-npm-webpack/app/index.js
```diff
@@ -0,0 +1,53 @@
...
+document.getElementById( "currency" ).innerHTML = currencyFormatter( 69900 );
+document.getElementById( "date" ).innerHTML = dateFormatter( new Date() );
+document.getElementById( "number" ).innerHTML = numberFormatter( 12345.6789 );
+document.getElementById( "relative-time" ).innerHTML = relativeTimeFormatter( 0 );
+
+// Messages.
+document.getElementById( "message-1" ).innerHTML = Globalize.formatMessage( "message-1", {
+	number: currencyFormatter( 69900 ),
+	currency: dateFormatter( new Date() ),
+	date: numberFormatter( 12345.6789 ),
```
sompylasar: Swap `date` and `currency` values or field names.

examples/app-npm-webpack/app/index.js
```diff
@@ -0,0 +1,53 @@
...
+
+// Refresh elapsed time
+setInterval(function() {
+	var elapsedTime = +( ( startTime - new Date() ) / 1000 ).toFixed( 0 );
+	document.getElementById( "date" ).innerHTML = dateFormatter( new Date() );
+	document.getElementById( "relative-time" ).innerHTML = relativeTimeFormatter( elapsedTime );
+	document.getElementById( "message-1" ).innerHTML = Globalize.formatMessage( "message-1", {
+		number: currencyFormatter( 69900 ),
+		currency: dateFormatter( new Date() ),
+		date: numberFormatter( 12345.6789 ),
```
sompylasar: Swap `date` and `currency` values or field names.

examples/app-npm-webpack/index-template.html
```diff
@@ -0,0 +1,82 @@
...
+		</tbody>
+	</table>
+	<p id="message-1">
+		An example of a message using mixed numbers "{number}", currencies "{currency}", dates "{date}", and relative time "{relativeTime}".
+	</p>
+	<p id="message-2">
+		An example of a message with pluralization support: 
+		{count, plural,
+			one {You have one remaining task}
+		  other {You have # remaining tasks}
```
sompylasar: Mixed indentation compared to the previous line.

examples/app-npm-webpack/index-template.html
```diff
@@ -0,0 +1,82 @@
...
+
+For displaying the application in a different locale, replace `en` with
+whatever other supported locale you want, e.g., `pt` (Portuguese).
+
+For supporting additional locales simultaneously and then having your
+application to change it dynamically, load the multiple files here. Then,
+use `Globalize.locale( <locale> )` in your application to dynamically set
+it.
+-->
+{%      } %}
```
sompylasar: Mixed indentation, see https://github.com/jquery/globalize/pull/481/files#diff-a2e501e11e8425260ad4ae42db5bddaaR59

examples/app-npm-webpack/webpack-config.js
```diff
@@ -0,0 +1,55 @@
+var webpack = require( "webpack" );
+var CommonsChunkPlugin = require( "webpack/lib/optimize/CommonsChunkPlugin" );
+var HtmlWebpackPlugin = require( "html-webpack-plugin" );
+var globalizePlugin = require( "globalize-webpack-plugin" );
```
sompylasar: `GlobalizePlugin` (uppercase) like the other plugins: `CommonsChunkPlugin`, `HtmlWebpackPlugin`.

examples/app-npm-webpack/app/index.js
rxaviers: Whops

rxaviers: 17089fb

examples/app-npm-webpack/index-template.html
rxaviers: e41c0a4

rxaviers: e41c0a4

examples/app-npm-webpack/index-template.html
rxaviers: 7204957

examples/app-npm-webpack/webpack-config.js
rxaviers: 014e900

examples/app-npm-webpack/app/index.js
rxaviers: c24dff9

examples/app-npm-webpack/app/index.js
rxaviers: :+1: should be tracked separately by #489

examples/app-npm-webpack/messages/en.json
rxaviers: For the record, Jörn has provided `de` translations.

rxaviers:
Pushed new Documentation content.

examples/app-npm-webpack/app/index.js
sompylasar: Yes, great!

README.md
```diff
@@ -299,6 +295,36 @@ For information about the Globalize Compiler CLI or its JavaScript API, see the
...
+### Examples
+
+The fastest and easiest way to use Globalize is by integrating it into your
+existing tools.
+
+- [Application example using webpack and npm](examples/app-npm-webpack/): easy to
+  get started, automated CLDR loading and precompilation for production, but
+  requires npm and webpack knowledge.
+
+Considering you're using a different tool than the one above. But, you're
```
kswedberg: I'd change this to:
> If you're using a tool other than webpack, but you're…

README.md
```diff
@@ -299,6 +295,36 @@ For information about the Globalize Compiler CLI or its JavaScript API, see the
...
+
+- [Basic example using AMD and bower](examples/amd-bower/): feeding Globalize on
+  CLDR is not completely transparent.
+- [Basic example using Node.js and npm](examples/node-npm/): feeding Globalize
+  on CLDR is not completely transparent.
+- [Basic Globalize Compiler example][]: shows how to use Globalize Compiler CLI.
+
+[Basic Globalize Compiler example]: examples/globalize-compiler/
+
+Considering you're using none of the tools above, but instead you're using
```
kswedberg: Change "Considering" to "If"

examples/app-npm-webpack/README.md
```diff
@@ -0,0 +1,74 @@
...
+```
+
+1. Start a server by running `npm start`, which uses webpack's live reload HMR
+(Hot Module Replacement). See `package.json` to understand the actual shell
+command that is used.
+1. Point your browser at `http://localhost:8080`. Note that your browser will
+automatically reload on any changes made to the application code (`app/*.js`
+files). Also note that for faster page reload, formatters are created
+dynamically and automatically by the [Globalize Webpack Plugin][].
+1. Note you can specify the development locale of your choice by setting up
```
kswedberg: by setting *the*

README.md
rxaviers: 92a4bc9

README.md
rxaviers: 92a4bc9

examples/app-npm-webpack/README.md
rxaviers: 92a4bc9

rxaviers:
Thank you all

examples/app-npm-webpack/messages/es.json
```diff
@@ -0,0 +1,17 @@
+{
+	"es": {
+		"intro-1": "Usa Globalize para internacionalizar tu aplicación.",
+		"number-label": "Número",
+		"currency-label": "Moneda",
+		"date-label": "Fecha",
+		"relative-time-label": "Tiempo Relativo",
+		"message-1": "Un ejemplo de mensaje usando números mixtos \"{number}\", monedas \"{currency}\", fechas \"{date}\", y tiempo relativo \"{relativeTime}\".",
+		"message-2": [
+			"Un ejemplo de mensaje con soporte de pluralizaciónt:",
```
unindented: Small typo, extra `t` at the end of `pluralización`.

juhamust:
I wonder what is the correct place to report the issues with `webpack-globalize`, so I'll a quick on here.

Feels that globalize compiler is trying to be too smart for its own good (too much optimisation?): Methods like `Globalize.numberFormatter` are considered undefined time to time. Assigning them to variable like `var numberFormatter = Globalize.numberFormatter` helps, but only sometimes. I haven't found the pattern, yet.

----

Also, evaluation may consider variables undefined, while they're not:

```javascript
// Angular translation filter
angular.module('app')
  .filter('t', function($filter) {
    return function (translationKey) {
      return Globalize.formatMessage(translationKey);
    };
  });
```

```javascript
// Compilation error
./node_modules/globalize-webpack-plugin/GlobalizeCompilerHelper.js:72
      throw e;
            ^
ReferenceError: translationKey is not defined
    at eval (eval at extractor (./node_modules/globalize-compiler/lib/extract.js:67:9), <anonymous>:3:36)
    at ./node_modules/globalize-compiler/lib/compile-extracts.js:54:23
    at Array.reduce (native)
```

If I change the translationKey into valid `translation-key` string, it works. But obviously always returns the same translation. And all this worked at some point :(

Any wise words to help compiler to make it work like expected?

examples/app-npm-webpack/messages/es.json
rxaviers: Thanks, d317164.

rxaviers:
Hi @juhamust,

The problem you're facing is because you're using a dynamic construction in your Globalize options and the compiler is static. It doesn't mean you cannot use dynamic patterns in your code, but you have to change your code like the below:

```js
var availableMessageFormatters = {
  foo: Globalize.messageFormatter(<foo>),
  bar: Globalize.messageFormatter(<bar>)
  ...
};

// Angular translation filter
angular.module('app')
  .filter('t', function($filter) {
    return function (translationKey) {
      return availableMessageFormatters[translationKey];
    };
  });
```

Note this is actually a good constrain for your code. Because, you could potentially run into a case where you are trying to format a message in production that you have never provided translation for.

Did I answer to your question? Would you change anything in our docs to make this clear?

PS: Anyway, if you need to report errors to `webpack-globalize`, do so at https://github.com/rxaviers/globalize-webpack-plugin.

juhamust:
@rxaviers thanks a lot for your prompt and informative response. That really took me surprised, so perhaps the documentation really could be improved about the topic.

Duh. I tried building multilingual lookup table, but it still has that dynamic translationKey variable there.
```javascript
// Build translation message lookup table
var messageLookupTable = {};
var messages = require('json!../common/locales/messages/en.json');

_.each(['fi', 'en', 'es', 'de'], function(languageCode) {
  messageLookupTable[languageCode] = {};
  // Using english translation keys as a reference
  _.each(messages['en'], function(msg, translationKey) {
    messageLookupTable[languageCode][translationKey] = Globalize.messageFormatter(translationKey);
  });
});
```

So I *really* need to generate JS versions from all of those JSON message files? That's something I would expect the compiler is doing on my behalf?

Thanks again.

rxaviers:
@juhamust, can you give a little more details on what you are trying to accomplish? Is this part of a frontend or backend application? Where are you going to use `messageLookupTable` later in your application? I'm asking, because Globalize is pretty flexible and knowing your application helps me to give a more accurate answer. Thx

juhamust:
@rxaviers [I responded in separate ticket](https://github.com/rxaviers/globalize-webpack-plugin/issues/6#issuecomment-145763411) instead of polluting this PR thread

rxaviers:
:+1: I've replied you there.

PR #484, opened 2015-08-22
Number: Refactor formatter to extract lots of variables

jzaefferer:
Ref https://github.com/jquery/globalize/pull/436#discussion_r36746515

### DIFF
--- a/src/number.js
+++ b/src/number.js
@@ -26,6 +26,33 @@ define([
 	numberFormatterFn, numberFormatProperties, numberNumberingSystem, numberParserFn,
 	numberParseProperties, numberPattern, numberSymbol, stringPad ) {
 
+function validateDigits( properties ) {
+	var minimumIntegerDigits = properties[ 2 ];
+	var minimumFractionDigits = properties[ 3 ];
+	var maximumFractionDigits = properties[ 4 ];
+
+	var minimumSignificantDigits = properties[ 5 ];
+	var maximumSignificantDigits = properties[ 6 ];
+
+	// Validate significant digit format properties
+	if ( !isNaN( minimumSignificantDigits * maximumSignificantDigits ) ) {
+		validateParameterRange( minimumSignificantDigits, "minimumSignificantDigits", 1, 21 );
+		validateParameterRange( maximumSignificantDigits, "maximumSignificantDigits",
+			minimumSignificantDigits, 21 );
+
+	} else if ( !isNaN( minimumSignificantDigits ) || !isNaN( maximumSignificantDigits ) ) {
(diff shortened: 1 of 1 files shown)


rxaviers:
@jzaefferer thanks for your change.

Although I'm not against it, I didn't understand the value of this change? All variables are still there, although they have been moved into a different function.

jzaefferer:
It makes the formatter function easier to understand, since there's a lot less variables in its scope.

scottgonzalez:
Seems good.

jzaefferer:
Rebased.

src/number.js
```diff
@@ -26,6 +26,33 @@ define([
 	numberFormatterFn, numberFormatProperties, numberNumberingSystem, numberParserFn,
 	numberParseProperties, numberPattern, numberSymbol, stringPad ) {
 
+function validateDigits( properties ) {
+	var minimumIntegerDigits = properties[ 2 ];
+	var minimumFractionDigits = properties[ 3 ];
+	var maximumFractionDigits = properties[ 4 ];
+
+	var minimumSignificantDigits = properties[ 5 ];
+	var maximumSignificantDigits = properties[ 6 ];
```
rxaviers: Coding style now allows `onevar: false`, which I personally like too. Although, the rest of the code still uses `onevar: true`. For consistency sake, I suggest this to be updated for `onevar` until #413 lands.

rxaviers:
I left one comment above. Other than that, let's get it landed. Thanks.

src/number.js
jzaefferer: We've been doing gradual migrations elsewhere. I see no point introducing new code with the old style just for consistency. Anything new shouldn't use `onevar` is easy enough to apply.

rxaviers: Where?

jzaefferer: For example:

https://github.com/jquery/jquery-ui/commit/803eaf29f732a68793b806356096fe849b9f470b#diff-29e4c930be69ff962e7aee3f71878229R144
https://github.com/jquery/qunit/commit/eaae69b822cb1410bd4c61eedd1b4a120757a5fe#diff-586f09c9d3a303fcf4f478bd0b1d5ca4R1

rxaviers: Well, the examples are from different projects. I wouldn't encourage having inconsistent code.

If it helps, feel free to cherry-pick https://github.com/rxaviers/globalize/commit/26f7e1049032eae3263796c7e6ed3ef15dbc4b31.

jzaefferer: I don't really mind, feel free to land this with your fixup commit.

Issue #252, opened 2014-05-19
Support formatting units, eg. secs, days


rxaviers:
PR #254

Format units such as seconds, minutes, days, weeks, and everything else supported by CLDR.

**Requires**
- #220 Plural

**API**

.formatUnit( value, unit [, options] );

Examples:
```javascript
.formatUnit( 1, "month" ); // 1 month
.formatUnit( 5, "week" ); // 5 weeks
.formatUnit( 30, "day" ); // 30 days
.formatUnit( 30, "day", { form: "narrow" }); // 30d
```

Value: Number

Unit: String, eg. "month", "day", etc.

Options:

- form ("long"|"short"|"narrow") used when tense = null to specify the unit form, eg. ("2 seconds", "2 secs", or "2s").

**Specs**

- [TR35 Unit Elements](http://www.unicode.org/reports/tr35/tr35-35/tr35-general.html#Unit_Elements)

**CLDR**

Units can be found on [main/<lang>/units.json](http://www.unicode.org/repos/cldr-aux/json/25/main/en/units.json).

**Ecma proposal**

- https://github.com/tc39/ecma402/issues/32

--
Fore more background: see #391.

arkihillel:
This is a must have as far as I'm concerned. Would be glad to help if needed. 
What I tried to do a week ago before realizing I couldn't with the current version of globalize might help you get a clear use case:

**Display week dates**
```javascript
function formatWeek(startDate, language) {
	var startDate = new Date(startDate);
	var i18nString = _('%s of %s to %s');
	var i18nWeek = Globalize.formatUnit( 1, "week" );
	var i18nFirstDay = Globalize.formatDate(startDate, {skeleton: 'yMd'});
	var i18nLastDay = Globalize.formatDate(startDate.setDate(startDate.getDate() + 6), language, {skeleton: 'yMd'});
	return util.format(i18nString, i18nWeek, i18nFirstDay, i18nLastDay);
}
```

rxaviers:
It was not clear to me whether the API specified here (and partially implemented by #252) is sufficient for you? If it is, you can help me to finish #252, so we can land it. Right now, it misses functional tests and docs. It got stalled due to lack of time/resource to work on it.

arkihillel:
#252 perfectly fits my needs. I'd be glad to help and will look into documenting it whenever I have free time.

rxaviers:
Awesome, feel free to add comments needing help. As soon as we get those items tick, it should be good to land on master.

[...]

PR #512, opened 2015-09-24
Add unit formatting module

alunny:
This is picking up @rxaviers's patch in #254 to solve #252.


alunny:
I think I can make it work by heavily cargo-culting relative-timestamp - new patch forthcoming.

alunny:
(I have spaces instead of tabs in many places - will update that when this is good to merge otherwise)

src/unit.js
```diff
@@ -0,0 +1,38 @@
...
+ * @value [Number]
+ *
+ * @unit [String]:
+ *
+ * @options [Object]
+ * - form: [String] "long", "short" (default), or "narrow".
+ *
+ * Format units such as seconds, minutes, days, weeks, etc.
+ */
+Globalize.prototype.formatUnit = function( value, unit, options ) {
```
rxaviers: Let's update this according to the below please. So, it's accessible via both static function (`Globalize.formatUnit`) and instance method (new `Globalize(locale).formatUnit`)

```js
Globalize.formatUnit =
Globalize.prototype.formatUnit = function( ...
```

src/unit.js
```diff
@@ -0,0 +1,38 @@
...
+  var pluralGenerator;
+
+	if ( typeof value !== "number" ) {
+		throw new Error( "Value is not a number" );
```
rxaviers: Let's reuse these data type validator https://github.com/jquery/globalize/blob/master/src/number.js#L151.

src/unit.js
```diff
@@ -0,0 +1,38 @@
...
+	}
+
+	if ( !unit ) {
+		throw new Error( "Missing unit" );
```
rxaviers: Let's reuse these data type validator https://github.com/jquery/globalize/blob/master/src/number.js#L150 (validateParameterPresence).

src/unit.js
```diff
@@ -0,0 +1,38 @@
...
+	}
+
+	pluralGenerator = this.pluralGenerator();
+
+	return unitFormat( value, unit, options, pluralGenerator, this.cldr, this );
+};
```
rxaviers: In general, `.formatUnit()` should be an alias for `.unitFormatter()`. Like `.formatNumber()` https://github.com/jquery/globalize/blob/master/src/number.js#L148-L154 is for `.numberFormatter()` https://github.com/jquery/globalize/blob/master/src/number.js#L107-L137

Note the process of setting up the formatter in general is somewhat an expensive operation. This process includes traversing CLDR data and cherry-picking the specific data (which I name `properties`) for the particular options used by the formatter. Therefore, the formatter generator should return a function that in turn allows to execute the formatting (i.e., allow user to setup once and execute several times for optimal performance; and to allow runtime compiled formatters). Please, just let me know on any questions.

src/unit/format.js
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * OBS:
+ *
+ * Unit Sequences are not implemented.
+ * http://www.unicode.org/reports/tr35/tr35-35/tr35-general.html#Unit_Sequences
+ *
+ * Duration Unit (for composed time unit durations) is not implemented.
+ * http://www.unicode.org/reports/tr35/tr35-35/tr35-general.html#durationUnit
+ */
+return function( value, unit, options, pluralGenerator, cldr, globalize ) {
```
rxaviers: Let's remove `globalize` as an argument. I know this comes from my previous branch. But, instead we should pass the plural formatter like https://github.com/jquery/globalize/blob/master/src/relative-time.js#L68. Because, it allows the runtime compiler to appropriately use a precompiled plural generator here.

src/unit/format.js
```diff
@@ -0,0 +1,61 @@
...
+	if ( ( /-per-|\// ).test( unit ) ) {
+
+		// "For the divisor, the 'one' plural category should be used, while for the
+		// dividend the appropriate plural form according the placeholder number
+		// should be used" UTS#35
+		//
+		// "There is a known problem with some languages in the long form in that
+		// the divisor should be inflected. This will probably require the future
+		// addition of a special 'divisor' form of units commonly used in the
+		// divisor." UTS#35
```
rxaviers: We should ensure this text (and therefore, the logic we're using here) is still valid. http://www.unicode.org/reports/tr35/tr35-general.html#Unit_Elements

src/unit/format.js
rxaviers: Just noticed we're already passing `pluralGenerator` above. But, below we're using `globalize` instance. Let's get the below fixed (and make the calling function passes the correct arguments).

test/unit/unit/format.js
```diff
@@ -0,0 +1,295 @@
...
+	);
+	assert.equal(
+    formatUnit( 1, "year", { form: "narrow" }, oneOrOtherPluralGenerator, cldr, globalize ),
+    "1y"
+	);
+	assert.equal(
+    formatUnit( 2, "year", { form: "narrow" }, oneOrOtherPluralGenerator, cldr, globalize ),
+    "2y"
+	);
+});
```
rxaviers: Let's simplify the whole assertions above by creating a custom assertion like https://github.com/jquery/globalize/blob/master/test/unit/date/format.js#L60-L71. This way we avoid boilerplate code.

src/unit.js
alunny: fixed on my local branch (commenting for my own tracking benefit)

src/unit.js
alunny: fixed on local branch

src/unit.js
alunny: fixed on local branch

test/unit/unit/format.js
alunny: fixed on local branch

src/unit/format.js
alunny: I think the compound unit logic is incorrect now - will fixup and write some tests.

src/unit/format.js
alunny: fixed in local branch

src/unit.js
alunny: I'm a little confused by this - is it right to say `unitFormatter` should be the only function that touches the cldr data, and that `src/unit/format.js` should operate on the already extracted content?

Here's an example of one approach I was looking at for this, let me know if it looks sound.
https://gist.github.com/alunny/c9777c4c171b986fad9f

rxaviers: Your gist diff seems perfect!

The point of keeping `src/unit/format.js` CLDR-free (and as light as possible) is that it goes in the runtime module. The other functions that setup and prepare/create the `properties` (e.g., `src/unit/get.js`) can be safely ignored in the runtime modules.

alunny: Sounds good - had to update the tests too, but now everything's passing.

src/unit.js
```diff
@@ -0,0 +1,57 @@
...
+	form = options.form || "long";
+	unitProperties = unitGet( unit, form, this.cldr );
+
+  return unitFormatterFn( unitProperties, this.pluralGenerator() );
+};
+
+return Globalize;
+
+});
+
```
rxaviers: Space vs tabs indentation is still used here. Tabs should be used for indentation.

rxaviers: There's an extra space in the bottom of the file.

rxaviers: There's an extra empty line in the bottom of the file. (EDITED)

src/unit.js
```diff
@@ -0,0 +1,57 @@
...
+	"./common/validate/parameter-presence",
+	"./common/validate/parameter-type/number",
+	"./common/validate/parameter-type/plain-object",
+	"./unit/format",
+	"./unit/formatter-fn",
+	"./unit/get",
+
+  "./plural"
+], function( Globalize, validateParameterPresence, validateParameterTypeNumber,
+  validateParameterTypePlainObject, unitFormat, unitFormatterFn, unitGet ) {
```
rxaviers: Unused `unitFormat`.

src/unit/get.js
```diff
@@ -0,0 +1,76 @@
...
+	return aux;
+}
+
+/**
+ * get( unit, form, cldr )
+ *
+ * @unit [String] The full type-unit name (eg. duration-second), or the short unit name
+ * (eg. second).
+ *
+ * FIXME
```
rxaviers: Let's replace `FIXME` by adding a description for `@form` and `@cldr` attributes.

src/unit.js
rxaviers: Wondering why `grunt jscs:source` didn't caught these...

src/unit/get.js
```diff
@@ -0,0 +1,76 @@
...
+			form,
+			category.length ? category + "-" + unit : unit
+		]);
+	});
+
+	// Rename keys s/unitPattern-count-//g.
+	ret = stripPluralGarbage( ret );
+
+	// Compound Unit, eg. "foot-per-second" or "foot/second".
+	if ( !ret && ( /-per-|\// ).test( unit ) ) {
```
rxaviers: I like allowing the use of `/` as well as `-per-`. But, `-per-` is what is going to be found on CLDR. So, in the beginning of this function we need `unit = unit.replace( /\// , "-per-" );`, otherwise `kilometer/hour` would not use the precomputed `kilometer-per-hour` form and it would get both recursively only.

src/unit/categories.js
```diff
@@ -0,0 +1,11 @@
+define(function() {
+
+/**
+ * categories()
+ *
+ * Return all unit categories.
+ */
+return [ "acceleration", "angle", "area", "duration", "length", "mass", "power", "pressure",
+"speed", "temperature", "volume" ];
```
rxaviers: An alternative to this approach is generating this dynamically once and keep it cached for further uses. We could keep it as is and improve later though.

test/unit/unit/format.js
```diff
@@ -0,0 +1,137 @@
...
+	assert.unitFormat( 2, "month", { form: "narrow" }, "2m" );
+	assert.unitFormat( 1, "year", { form: "narrow" }, "1y" );
+	assert.unitFormat( 2, "year", { form: "narrow" }, "2y" );
+});
+
+QUnit.test( "Compound form (long)", function ( assert ) {
+	assert.unitFormat( 1, "speed-mile-per-hour", {}, "1 mile per hour" );
+	assert.unitFormat( 100, "speed-mile-per-hour", {}, "100 miles per hour" );
+	assert.unitFormat( 1, "consumption-mile-per-gallon", {}, "1 mile per gallon" );
+	assert.unitFormat( 100, "consumption-mile-per-gallon", {}, "100 miles per gallon" );
```
rxaviers: Let's insert a unit without category for completeness, e.g., `mile-per-hour` only.

test/unit/unit/format.js
```diff
@@ -0,0 +1,137 @@
...
+	assert.unitFormat( 1, "hour", { form: "narrow" }, "1h" );
+	assert.unitFormat( 2, "hour", { form: "narrow" }, "2h" );
+	assert.unitFormat( 1, "day", { form: "narrow" }, "1d" );
+	assert.unitFormat( 2, "day", { form: "narrow" }, "2d" );
+	assert.unitFormat( 1, "week", { form: "narrow" }, "1w" );
+	assert.unitFormat( 2, "week", { form: "narrow" }, "2w" );
+	assert.unitFormat( 1, "month", { form: "narrow" }, "1m" );
+	assert.unitFormat( 2, "month", { form: "narrow" }, "2m" );
+	assert.unitFormat( 1, "year", { form: "narrow" }, "1y" );
+	assert.unitFormat( 2, "year", { form: "narrow" }, "2y" );
```
rxaviers: Let's insert a unit with its category for completeness, e.g., `speed-mile`.

test/unit/unit/format.js
rxaviers: Also, let's include a compound without a precomputed value, e.g., `megabyte-per-second`.

rxaviers:
Great work so far!! I have left a couple of comments above. Other than that, I also miss these points:
- [ ] Update docs (refs [1](https://github.com/jquery/globalize/blob/master/README.md#pick-the-modules-you-need), [2](https://github.com/jquery/globalize/blob/master/README.md#2-cldr-content), [3](https://github.com/jquery/globalize/blob/master/README.md#relative-time-module));
- [ ] Build unit module (refs [1](https://github.com/jquery/globalize/blob/master/Gruntfile.js#L361-L371), [2](https://github.com/jquery/globalize/blob/master/Gruntfile.js#L537));
- [ ] Create functional tests (ref [1](https://github.com/jquery/globalize/tree/master/test/functional/relative-time));
- [ ] Build runtime module (refs [1](https://github.com/jquery/globalize/blob/master/Gruntfile.js#L461-L476), [2](https://github.com/jquery/globalize/blob/master/Gruntfile.js#L547))

Just let me know on any questions and thanks for the great work!

alunny:
Thanks for the great feedback - working on cleaning up this branch now :)

src/unit/categories.js
alunny: Modified this slightly to add "digital" category. Are there other places in the codebase where you're caching once based on the cldr data?

test/unit/unit/format.js
```diff
@@ -0,0 +1,152 @@
...
+
+	form = options.form || "long";
+	unitProperties = unitGet( unit, form, cldr );
+	compoundUnitPattern = unitGet( "per", form, cldr ).compoundUnitPattern;
+
+	this.equal(
+		formatUnit( value, unitProperties, oneOrOtherPluralGenerator, compoundUnitPattern ),
+		expected
+	);
+};
```
alunny: I'm not really happy with the unitFormat function - it's a bit too coupled with `Globalize.formatUnit`. Suggestions for refactoring are welcome.

rxaviers:
> > An alternative to this approach is generating this dynamically once and keep it cached for further uses. Although, this could remain untouched for now and get improved later.
> Modified this slightly to add "digital" category. Are there other places in the codebase where you're caching once based on the cldr data?

Ok and nope we have nothing similar in our codebase currently. Actually, I dislike to embed any sort of data in the code unless absolutely needed like in this case and a couple others: [[1](https://github.com/jquery/globalize/blob/master/src/number/numbering-system-digits-map.js#L17)] and [[2](https://github.com/jquery/globalize/blob/master/src/number/symbol/name.js#L4-L10)].

src/unit.js
```diff
@@ -0,0 +1,56 @@
...
+	var compoundUnitPattern, form, unitProperties;
+
+	validateParameterPresence( unit, "unit" );
+	validateParameterTypePlainObject( options, "options" );
+
+	form = options.form || "long";
+	compoundUnitPattern = unitGet( "per", form, this.cldr ).compoundUnitPattern;
+	unitProperties = unitGet( unit, form, this.cldr );
+
+	return unitFormatterFn( unitProperties, this.pluralGenerator(), compoundUnitPattern );
```
rxaviers: I'd like to suggest two changes:

1) Simplify how you get `compoundUnitPattern` (I think reusing `unitGet` for that is an overkill and kinda confusing)

```diff
- compoundUnitPattern = unitGet( "per", form, this.cldr ).compoundUnitPattern
+ compoundUnitPattern: this.cldr.main(["units", form, "per/compoundUnitPattern"])
```

2) Collapse properties:

```diff
-	var compoundUnitPattern, form, unitProperties;
+	var form, properties;
 
 	validateParameterPresence( unit, "unit" );
 	validateParameterTypePlainObject( options, "options" );
 
 	form = options.form || "long";
-	compoundUnitPattern = unitGet( "per", form, this.cldr ).compoundUnitPattern;
-	unitProperties = unitGet( unit, form, this.cldr );
+	properties = {
+		compoundUnitPattern: this.cldr.main(["units", form, "per/compoundUnitPattern"])
+		unit: unitGet( unit, form, this.cldr )
+	}
 
-	return unitFormatterFn( unitProperties, this.pluralGenerator(), compoundUnitPattern );
+	return unitFormatterFn( properties, this.pluralGenerator() );
```

src/unit/get.js
```diff
@@ -0,0 +1,82 @@
...
+ *     "other": "{0} feet" },
+ *   { "one": "{0} second",
+ *     "other": "{0} seconds" } ]
+ *
+ * Or undefined in case the unit (or a unit of the compound-unit) doesn't exist.
+ */
+var get = function( unit, form, cldr ) {
+	var ret;
+
+	// Ensure that we get the 'precomputed' form, if present.
```
rxaviers: Let's include we're treating `/` like `-per`, something like "Treat `/` like `-per-` and ensure that we get the 'precomputed' form, if present."

src/unit/get.js
```diff
@@ -0,0 +1,82 @@
...
+	unit = unit.replace( /\//, "-per-" );
+
+	// Get unit or <type>-unit (eg. "duration-second").
```
rxaviers: s/type/category

```diff
-	// Get unit or <type>-unit (eg. "duration-second").
+	// Get unit or <category>-unit (eg. "duration-second").
```

src/unit.js
rxaviers: Perhaps moving the code below in its own function, like `unit/properties` might avoid duplicate code in tests:
```
properties = {
    compoundUnitPattern: this.cldr.main(["units", form, "per/compoundUnitPattern"])
    unit: unitGet( unit, form, this.cldr )
}
```

test/unit/unit/format.js
rxaviers: Feel free to take or not this suggestion. But, I would remove the `"long"` default for `form` (in here, unit tests) and I would collapse both unit and compountUnitPattern in a single function `unit/properties.js` to avoid duplicating changes here and there.

rxaviers:
I've removed the below from the description since it's outdated. Please, correct me if I'm wrong.

> The tests don't fully pass - line 56 in `src/unit/format.js` is causing problems:
```js
    return globalize.formatPlural( value, ret );
```

> I can dig in further, but if @rxaviers or someone else on the Globalize knows what API should be called now, that would be helpful.

test/unit/unit/format.js
alunny: I think that's a good idea, thanks.

alunny:
Yeah you're right to edit the description, the tests all pass now.

alunny:
I've added the module builds, runtime support, functional tests, and docs now. Phew!

Gruntfile.js
```diff
@@ -378,6 +378,22 @@ module.exports = function( grunt ) {
 							}
 						},
 						{
+							name: "globalize.unit",
```
jzaefferer: GitHub has some kind of issue with this file. Looking at the source locally, I also get messed up syntax highlighting in my editor (SublimeText). Looks like an issue with the regex on line 192, completely unrelated to this PR though. Somehow this helps:
```diff
diff --git a/Gruntfile.js b/Gruntfile.js
index ce458fc..60819b1 100644
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -189,7 +189,7 @@ module.exports = function( grunt ) {
                             // Remove browserify wrappers.
                             .replace( /^\(function\(f\){if\(typeof exports==="object"&&type.*/, "" )
                             .replace( "},{}],2:[function(require,module,exports){", "" )
-                            .replace( /},{"\.\/messageformat-parser":1,"make-plural\/plural.*/, "" )
+                            .replace( /},{"\.\/messageformat-parser\":1,"make-plural\/plural.*/, "" )
                             .replace( /},{}\]},{},\[2\]\)\(2\)[\s\S]*?$/, "" )
 
                             // Set `MessageFormat.plurals` and remove `make-plural/plurals`
```

README.md
```diff
@@ -571,6 +574,29 @@ handle dependencies and CLDR loading manually yourself.
 
   Alias for `.relativeTimeFormatter( unit, options )( value )`.
 
+## Unit module
+
+-- **`.unitFormatter( unit, options )`**
```
jzaefferer: The two dashes at the start of the line should be a single dash, to match the formatting from other API entries.

doc/api/unit/unit-formatter.md
```diff
@@ -0,0 +1,56 @@
...
+formatter = Globalize.unitFormatter( "month", { form: "long" } );
+
+formatter( 1 );
+// > "1 month"
+
+formatter( 3 );
+// > "3 months"
+```
+
+You can use the instance method `.relativeTimeFormatter()`, which uses the instance locale.
```
jzaefferer: Can copy&paste? Seems like this should refer to `formatUnit`.

doc/api/unit/unit-formatter.md
```diff
@@ -0,0 +1,56 @@
...
+### Parameters
+
+**unit**
+
+String value indicating the unit to be formatted. eg. "day", "week", "month", etc.
+Could also be a compound unit, eg. "mile-per-hour" or "mile/hour"
+
+**options**
+
+- form: [String] eg. "long", "short" or "narrow".
```
jzaefferer: There are no options for formatting the number itself?

src/unit/categories.js
```diff
@@ -0,0 +1,11 @@
+define(function() {
+
+/**
+ * categories()
+ *
+ * Return all unit categories.
+ */
+return [ "acceleration", "angle", "area", "digital", "duration", "length", "mass", "power",
```
jzaefferer: This is used in exactly one place, does that really justify a separate file?

src/unit/format.js
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * format( value, unit, pluralGenerator )
+ *
+ * @value [Number]
+ *
+ * @unitProperies [Object]: localized unit data from cldr.
+ *
+ * @pluralGenerator [Object]: A pluralGenerator from Globalize.pluralGenerator.
+ *
+ * TODO pass along numberFormatter
```
jzaefferer: Is this tracked in a separate issue?

doc/api/unit/unit-formatter.md
rxaviers: Shoot. Good point.

I would add a `numberFormatter` option that takes a formatter as option.

src/unit/categories.js
rxaviers: Yes, it helps to make the build easier. This could change though when the source code uses ES6.

src/unit/format.js
rxaviers: Oh! So, here it is. I am ok to tracked this as a separate issue. I will let @alunny answer if he is comfortable including this in this PR or to leave it to a subsequent one.

Gruntfile.js
rxaviers: @jzaefferer can you file a separate ticket for this please?

src/unit/format.js
rxaviers: Let me correct myself... The number needs to be formatted using `numberFormatter()` (instead of being passed directly to the formatMessage as is below) in this PR, but we can defer allowing users to pass a different `numberFormatter()` as an option to a subsequent PR.

Use [relativeTimeFormatter](https://github.com/jquery/globalize/blob/master/src/relative-time.js#L67) as reference. But, note that I suggested a minor different thing than what relativeTimeFormatter does. It passes along its own options to the numberFormatter, which raises a new point: both should behave analogously.

I believe allowing a user to pass the formatter as an option is better than mixing options of both unit formatter and number formatter. What do you think?

jzaefferer: Since the unit formatter needs to access the actual number, passing a custom number formatter sounds fine to me. Would be good to include in this PR.

alunny: I don't think it's too much trouble to pass a custom number formatter, I'll take a look at that today.

Gruntfile.js
rxaviers: for the record, #527 (fixed by #529)

test/functional/unit/format-unit.js
```diff
@@ -0,0 +1,135 @@
...
+			Globalize.formatUnit( 0, invalidValue );
+		};
+	});
+});
+
+QUnit.test( "should validate options argument presence", function( assert ) {
+	util.assertParameterPresence( assert, "options", function( ) {
+		Globalize.formatUnit( 1, "day" );
+	});
+});
```
rxaviers: Actually options should be optional. :stuck_out_tongue_closed_eyes:

rxaviers: I'm working on ti.

rxaviers:
Awesome! Rebased and merged. Thx @alunny

rxaviers:
Published in 1.1.0-rc.4

alunny:
Great, thanks @rxaviers!

jzaefferer:
:clap: :clap: :clap:

PR #515, opened 2015-09-29
Docs: Improve documentation for raw date formatter option

### LINKED ISSUE #500: Parse date with custom format
I need to parse date in format "year-month-day", for example "2011-01-01".
How can I do that?

https://github.com/jquery/globalize/blob/master/doc/api/date/date-formatter.md

Describes a lot of options but noone seems to support custom separators for parsing.
I expected to use `skeleton` but it turns out to be of the same rigid presets.

This is very strange unless I missed something.
There is no "standard" separator for date in Russian locale, for example.
We use "-" as frequent as "/".

jzaefferer:
Fixes #500 

There, I fixed it.

### DIFF
--- a/doc/api/date/date-formatter.md
+++ b/doc/api/date/date-formatter.md
@@ -50,8 +50,8 @@ A JSON object including one of the following.
 > **raw**
 >
 > String value indicating a machine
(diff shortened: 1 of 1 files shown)


rxaviers:
:+1: thanks

Issue #517, opened 2015-10-01
Decimal separator ignored at the beginning of parsed value, e.g. ".5" is parsed as 5


kolman:
Decimal separator is ignored when the parsed value does not have leading zero:
```
Globalize('en').parseNumber('.5');
// => 5
parseFloat('.5');
// => 0.5
```

Mottie:
Duplicate of #443

rxaviers:
Thank you @Mottie and @kolman.

Issue #454, opened 2015-06-18
Issue with numberParser when price is .5


vpateliya:
Hi I am facing an issue when we pass .5 the number returns 5.00 instead of 0.50. Can we fix that?
var price = .5
console.log(numberParser(price))

kborchers:
Thanks, this is already being tracked in #443

Issue #353, opened 2014-12-02
Number: parse literals


rxaviers:
Handle literals when parsing numbers. E.g.,

```
Globalize.parseNumber("-$1,214.12", { pattern: "'$'#,##0.##" });
```

Test:
- [x] #443
- [x] #457
- [x] #46
- [x] #492
- [x] #587
- [x] #644

References:
- http://www.unicode.org/reports/tr35/tr35-numbers.html#Special_Pattern_Characters
- analogous to `src/date/token*` and `src/date/parse*`. Look at https://github.com/jquery/globalize/blob/0b85f34dc1a47bc2b9c6f8bbed15a7b7b8f4752f/src/date/tokenizer.js#L369-L376.
  - http://www.unicode.org/reports/tr35/tr35-dates.html#Date_Format_Patterns

lijunle:
Hi, @rxaviers This bug is open on year 2014. Any updates on it?

rxaviers:
The project is out of resources and is relying on external contribution to get this feature implemented.

Having said that, this feature is still open to be included. If you're willing to contribute it, I can provide further any guidance you may need to get up to speed. Thanks.

lijunle:
@rxaviers Could you provide some resources and ideas about how to implement this? Besides, do you have some workaround on this?

rxaviers:
The workaround is to do any parsing at user land and pass a number only to numberParser, which isn't great.

I've updated one resource link in the description. Please, let me know what you think after going through them.

The files that will need change is `src/number/parse.js`. The changes should be analogous to `src/date/{parse, tokenizer}*`.

Please, just let me know if you have any further questions.

rxaviers:
Edit: One file that will need change is `src/number/parse.js`. The changes should be analogous to `src/date/{parse, tokenizer}*`.

rxaviers:
PR #644 in place :)

rxaviers:
PR #656 in place :) (EDITED)

PR #518, opened 2015-10-02
Number: Add padding zero to leading decimal

### LINKED ISSUE #443: numberParser incorrect when no integer is included
I set up [this demo](http://jsfiddle.net/Mottie/gf3nu9q6/) using only a partial set of data, so maybe I'm doing it wrong...

```js
Globalize.numberParser()('.44');
// result = 44 (no decimal point)
Globalize.numberParser()('0.44');
// result = .44
```

Mottie:
Fixes #443
Ref #353
Ref #454
Ref #517 

I know this isn't an ideal solution, but adding a padding zero is way easier than trying to modify the regular expression, then adapting the code that depends on the matches.

New unit tests were added, and all passing.

### DIFF
--- a/test/unit/number/parse.js
+++ b/test/unit/number/parse.js
@@ -5,6 +5,7 @@ define([
 	"json!cldr-data/main/ar/numbers.json",
 	"json!cldr-data/main/en/numbers.json",
 	"json!cldr-data/main/es/numbers.json",
+	"json!cldr-data/main/pt/numbers.json",
 	"json!cldr-data/main/ru/numbers.json",
 	"json!cldr-data/main/sv/numbers.json",
 	"json!cldr-data/main/zh/numbers.json",
@@ -13,15 +14,16 @@ define([
 
 	"cldr/event",
 	"cldr/supplemental"
-], function( Cldr, parse, properties, arNumbers, enNumbers, esNumbers, ruNumbers, svNumbers,
-	zhNumbers, likelySubtags, numberingSystems ) {
+], function( Cldr, parse, properties, arNumbers, enNumbers, esNumbers, ptNumbers, ruNumbers,
+	svNumbers, zhNumbers, likelySubtags, numberingSystems ) {
 
-var ar, en, es, ru, sv, zh;
+var ar, en, es, pt, ru, sv, zh;
 
 Cldr.load(
 	arNumbers,
 	enNumbers,
 	esNumbers,
+	ptNumbers,
 	ruNumbers,
 	svNumbers,
 	zhNumbers,
@@ -32,6 +34,7 @@ Cldr.load(
 ar = new Cldr( "ar" );
 en = new Cldr( "en" );
 es = new Cldr( "es" );
+pt = new Cldr( "pt" );
 ru = new Cldr( "sv" );
 sv = new Cldr( "sv" );
 zh = new Cldr( "zh-u-nu-native" );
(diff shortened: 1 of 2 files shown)


Mottie:
Hmm, I signed the CLA... three times! The first time I used "Rob Garrison", but that didn't match the git `user.name`... so, the last two times I used "Mottie", but it is still not reporting the CLA as signed. Anyway, "I AGREE"!

jzaefferer:
Looks like the issue here is that the CLA checker expects names to have at least one space (like "Rob Garrison"), but you've used "Mottie" as your git author name. Amending the commit to change the author should fix it, since it then matches your first signature. We can also take care of that when landing your patch.

test/unit/number/parse.js
```diff
@@ -78,13 +78,18 @@ QUnit.test( "should parse zero-padded decimals", function( assert ) {
 	assert.equal( parse( "0.10", properties( "0.00", en ) ), 0.1 );
 });
 
+QUnit.test( "should parse non-padded decimals", function( assert ) {
+	assert.equal( parse( ".14159", properties( "0.0", en ) ), 0.14159 );
+	assert.equal( parse( ".752", properties( "0.0", en ) ), 0.752 );
+});
```
rxaviers: Can you add tests for locales that use:
1. different decimal point (like Portuguese)?
1. different orientation (like Arabic)?

PS: You could format the number in these locales to know what the output would be and remove the `0`. :)

Mottie:
Yeah, I updated my config user.name and did the forced update per instructions, then just pushed a regular commit. Still no luck :disappointed:

test/unit/number/parse.js
Mottie: LOL I would if I knew what the other formats looked like. Should I just copy them from other tests?

rxaviers:
@Mottie, thanks for your patch. I've added one comment above about tests. Other than that, it looks great! Thanks!

test/unit/number/parse.js
rxaviers: You can `Globalize("ar").formatNumber(0.14159);` :smile: (considering you have loaded data for this locale)

Mottie: Ok, something weird is going on... when I use the above code, I get this result `٠٫١٤٢`... I removed the `٠` assuming it's the zero.

Then I plug it into `Globalize( "ar" ).numberParser()( "٫١٤٢" )` and the result becomes `0.142`. I have no idea why it is rounding the number.

If I use `Globalize( "pt" ).numberParser()(',14159')`, the result is `0.14159` as expected.

rxaviers: Use `maximumFractionDigits` option when formatting to increase the default, which is `3`. For example:
```js
Globalize("ar").formatNumber(0.14159, {maximumFractionDigits: 10})
// > '٠٫١٤١٥٩'
```

Try using `٫١٤١٥٩`...

Mottie:
I still don't know why it's saying I haven't signed the CLA. Maybe I'm in the database more than once, and it's only catching the first instance?

rxaviers:
@Mottie, excellent. Thanks for including the tests. Don't worry about the CLA check for this PR, because I can squash using the name from your last commit, which is in the CLA database as @jzaefferer mentioned.

rxaviers:
Closed by 07fac0f60381c87ad1735b740be0fb30cddc0b66 and published in 1.1.0-rc.3.

PR #525, opened 2015-10-07
Build: Bump requirejs version

apsdehal:
This update is also required as in to support intern. You can see the exact issue theintern/intern#517


rxaviers:
Thx

Issue #548, opened 2015-10-27
package.json: peerDependency no longer works


Mottie:
While performing an `npm install` the following message appears:

> $ npm install
> npm WARN peerDependencies The peer dependency cldr-data included from globalize will no
> npm WARN peerDependencies longer be automatically installed to fulfill the peerDependency
> npm WARN peerDependencies in npm 3+. Your application will need to depend on it explicitly.

> \> cldr-data@28.0.2 install D:\Repos\globalize\node_modules\cldr-data
> \> node install.js

> Warning: Skipping to download CLDR data, because `cldr-data` is a peer dependency. If you want it to be downloaded, make sure it's listed under `dependencies` or `devDependencies` of the `package.json`of your application.

rxaviers:
Hi, sorry for the delay. Please, could you provide context of the problem? I assume that when installing Globalize solely using npm intall, you `npm install globalize cldr-data` and when using it as a dependency of a project, you define both `globalize` and `cldr-data` as dependencies. In both cases, there's no `npm` complain.

Mottie:
I see those issues if I use GitHub Desktop (button from code page ![2016-01-19 15_11_56-jquery_globalize - javascript](https://cloud.githubusercontent.com/assets/136959/12432143/29ef7d90-bebf-11e5-8758-af5fd545dfd8.png)) to download the repository, or download a zip and save it to my drive.

Then, I use `npm install` to get npm to automatically grab all the needed dependencies, it won't include CLDR-data because it is a peer dependency. So, it requires an extra step of typing `npm install cldr-data`.

Trust me, it was a pain figuring all that out when you first start learning now to use npm.

rxaviers:
Initially, I assumed you were talking about it from a user perspective, but I understand now that you are doing so from a contributor perspective developing Globalize. Now, I understand and confirm your issue. These warnings show up on npm@2 and npm@3 and it happens because cldr-data is a peer dependency (so, the declaration is correct), but we don't use it as devDependencies (which npm assumes we would).

I am open to suggestions on how to improve this for now. I think we'll only get rid of this problem when we actually start using the npm packages (instead of the bower ones.), which might happen when we update tests #450.

dvlsg:
This also caused me a bit of a headache. Side note -- for me, even when executing `npm install --save cldr-data@latest` after a regular `npm install`, it didn't download the cldr data automatically. Same error as above.

[...]

PR #549, opened 2015-10-27
package.json: remove peerDependencies

Mottie:
Fixes gh-548

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,9 @@
     "url": "http://github.com/jquery/globalize/issues"
   },
   "dependencies": {
-    "cldrjs": "0.4.3",
+    "cldrjs": "0.4.3"
+  },
+  "peerDependencies": {
     "cldr-data": ">=25"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -72,9 +72,7 @@
     "url": "http://github.com/jquery/globalize/issues"
   },
   "dependencies": {
-    "cldrjs": "0.4.3"
-  },
-  "peerDependencies": {
+    "cldrjs": "0.4.3",
```
jzaefferer: This seems like the wrong fix. We want to keep the peerDependency, to allow users to install the version of their choice. We also want to install it when testing this project, so we should keep it in `peerDependencies`, while also adding it to `devDependencies`.

Mottie: After moving `"cldr-data": ">=25"` to `devDependencies` and adding a copy to `peerDependencies`, the warning message no longer appears. Do you want me to update the PR?

jzaefferer: Yes please

rxaviers:
Hi @Mottie, a good news is that we removed cldr-data as a peer dependency (details in #638). Probably, longer than you expected though. Thanks for bringing it to us.

PR #559, opened 2015-12-01
Support for RBNF (algorithmic numeric formatting) as part of numeric formatting in module number 

ashensis:
As per was discussed in:
General purpose API for converting regular numerals to Hebrew ones #537
this PR contains both general framework for adding algorithmic numeric formatting functions
and implementation for Hebrew algorithmic numeric formatter

### DIFF
--- a/examples/app-npm-webpack/package.json
+++ b/examples/app-npm-webpack/package.json
@@ -2,6 +2,7 @@
   "private": true,
   "devDependencies": {
     "cldr-data": ">=25",
+    "globalize-webpack-plugin": "0.3.x",
     "html-webpack-plugin": "^1.1.0",
     "nopt": "^3.0.3",
     "webpack": "^1.9.0",


rxaviers:
Thanks for submitting this PR.

I expected an implementation that (a) would handle RBNF rules http://unicode.org/repos/cldr/trunk/common/rbnf/root.xml, note I'm using xml as reference here, but the JSON bindings should be used as soon as it's [ready](https://github.com/jquery/globalize/issues/537#issuecomment-160694831), and (b) would follow what is described in [1] and [2] below. But, unless I'm wrong all I see is a Hebrew specific implementation. The implementation should work for any algorithmic numbering system given its RBNF rules.

Does that make sense? Thanks

1: http://www.icu-project.org/apiref/icu4c/classRuleBasedNumberFormat.html
2: http://www.unicode.org/reports/tr35/tr35-numbers.html#Rule-Based_Number_Formatting

ashensis:
Thank you Rafael, now I am getting my head around this (hopefully)
So you expected some internal general function that would accept RULES JSON (analogous to http://unicode.org/repos/cldr/trunk/common/rbnf/root.xml) when it gets available and will apply contained in this JSON rules, according to current language, in order to format/parse the number passed as a parameter to this function.
If I understand you correctly, this 'rules' JSON (or its part) should be passed into Globalize on initialization, alongside with the rest of CLDR JSON stuff been used currently.
Is my understanding correct?
If the answer on my question is positive, currently we have nothing to do until RBNF JSON will be available.

rxaviers:
Your understanding is precisely correct.

Until the RBNF JSON is available, it might be possible to get the implementation ready except for the JSON paths where to find the information from. I mean, we know the content given the XMLs, but we don't know the JSON paths for such contents which will be revealed after the bindings are defined.

Issue #576, opened 2016-01-12
Strict parsing dates?


hippich:
I am trying to use `parseDate` to parse dates supplied by user, but I need to catch incorrect dates. While in some cases it is seems to be possible, in others, like date '1/12/123' - is not, due `yy` in en_US locale matching `12` and silently dropping `3`...

Is there any way force strict matching of the value provided?

rxaviers:
Right now there's no way to force strict matching. I think this feature would be welcome.

PR #577, opened 2016-01-12
Date: Add 'strict' date parsing option.

hippich:
- Set `strict` to `true` to force strict parsing of dates. This will prevent
   parsing date `"01/01/2012"` when using pattern `"M/d/yy"`. Without `strict`
   this date will be parsed as `"01/01/20"`.

 Ref #576

### DIFF
--- a/test/unit/date/parse.js
+++ b/test/unit/date/parse.js
@@ -18,11 +18,11 @@ define([
 
 var cldr, date1, date2, FakeDate, midnight;
 
-function assertParse( assert, stringDate, pattern, cldr, date, strict ) {
+function assertParse( assert, stringDate, pattern, cldr, date ) {
 	var tokenizerProperties, tokens;
 
 	tokenizerProperties = numberTokenizerProperties( pattern, cldr );
-	tokens = tokenizer( stringDate, simpleNumberParser, tokenizerProperties, strict );
+	tokens = tokenizer( stringDate, simpleNumberParser, tokenizerProperties );
 
 	assert.deepEqual( parse( stringDate, tokens, parseProperties( cldr ) ), date );
 }
@@ -469,12 +469,6 @@ QUnit.test( "should parse milliseconds in a day (A+)", function( assert ) {
 	assertParse( assert, "63307369", "AAA", cldr, date1 );
 	assertParse( assert, "633073690", "AAAA", cldr, date1 );
 	assertParse( assert, "6330736900", "AAAAA", cldr, date1 );
-});
-
-QUnit.test( "should allow strict parsing of the date", function( assert ) {
-	date1 = new Date(2001,0,1);
-	assertParse( assert, "1/1/01", "M/d/yy", cldr, date1, true );
-	assertParse( assert, "1/1/2001", "M/d/yy", cldr, null, true );
 });
 
 /**
--- a/test/unit/date/parse.js
+++ b/test/unit/date/parse.js
@@ -18,11 +18,11 @@ define([
 
 var cldr, date1, date2, FakeDate, midnight;
 
-function assertParse( assert, stringDate, pattern, cldr, date ) {
+function assertParse( assert, stringDate, pattern, cldr, date, strict ) {
 	var tokenizerProperties, tokens;
 
 	tokenizerProperties = numberTokenizerProperties( pattern, cldr );
-	tokens = tokenizer( stringDate, simpleNumberParser, tokenizerProperties );
+	tokens = tokenizer( stringDate, simpleNumberParser, tokenizerProperties, strict );
 
 	assert.deepEqual( parse( stringDate, tokens, parseProperties( cldr ) ), date );
 }
@@ -469,6 +469,12 @@ QUnit.test( "should parse milliseconds in a day (A+)", function( assert ) {
 	assertParse( assert, "63307369", "AAA", cldr, date1 );
 	assertParse( assert, "633073690", "AAAA", cldr, date1 );
 	assertParse( assert, "6330736900", "AAAAA", cldr, date1 );
+});
+
+QUnit.test( "should allow strict parsing of the date", function( assert ) {
+	date1 = new Date(2001,0,1);
+	assertParse( assert, "1/1/01", "M/d/yy", cldr, date1, true );
+	assertParse( assert, "1/1/2001", "M/d/yy", cldr, null, true );
 });
 
 /**
--- a/doc/api/date/date-parser.md
+++ b/doc/api/date/date-parser.md
@@ -1,4 +1,4 @@
-## .dateParser( [options] ) ➜ function( value )
+## .dateParser( [options, strict] ) ➜ function( value )
 
 Return a function that parses a string representing a date into a JavaScript
 Date object according to the given `options`. The default parsing assumes numeric year, month, and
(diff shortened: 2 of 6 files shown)


jzaefferer:
Wouldn't it make more sense to add an option called `strict`, instead of adding another argument?

hippich:
For me `options` felt like options of format, rather than how this format should be applied. But I certainly not against it.. Any1 else to vote?

rxaviers:
Excellent @hippich! It's great to see your solution was succinct and that you have also added tests and documentation.

About `strict`, I also think it would be ok (and better) making it another `option`. I have some other comments that I am going to make inline in your code.

test/unit/date/parse.js
```diff
@@ -471,6 +471,12 @@ QUnit.test( "should parse milliseconds in a day (A+)", function( assert ) {
 	assertParse( assert, "6330736900", "AAAAA", cldr, date1 );
 });
 
+QUnit.test( "should allow strict parsing of the date", function( assert ) {
+	date1 = new Date(2001,0,1);
+	assertParse( assert, "1/1/01", "M/d/yy", cldr, date1, true );
+	assertParse( assert, "1/1/2001", "M/d/yy", cldr, null, true );
+});
```
rxaviers: Let's include `strict` tests for additional fields. I have tested it for `M`, but it falsely passes.

```js
assertParse( assert, "01/1/01", "M/d/yy", cldr, null, true );
// Should return null because of the padded month, but it returns date1.
```

So:

 1. Fix solution for `M` case above.
 1. What other fields can we test for? Only numerical? Numerical timezones? Any non-numerical ones?

rxaviers: Also note that this test is placed under the **second** section. Let's move it under the **year** section and place in here a `strict` test for some second fields.

rxaviers:
Given your fix is pretty much a change in the `src/date/tokenizer.js` file, let's include a couple of unit tests at `test/unit/date/tokenizer.js`. Actually, I think this test will help reveal the cause for the bug pointed https://github.com/jquery/globalize/pull/577#discussion_r49515832.

rxaviers:
Clarifying one of my comments: About strict, I also think it would be ok (and better) making it another key of `option`.

test/unit/date/parse.js
hippich: It appears to me that `M` will match both padded and not padded month. My intention was to avoid 2016 year being parsed as 2020 year when yy matches first two digits of 20116.

hippich: By that I mean it appears to be correct to parse `01` under `M` pattern. What do you think @rxaviers ?

rxaviers: > My intention was to avoid 2016 year being parsed as 2020 year when yy matches first two digits of 20116.

I understand, but supporting `strict` should work for other fields as well.

> It appears to me that M will match both padded and not padded month... By that I mean it appears to be correct to parse `01` under `M` pattern

Formatting `January` with `M` in `en` will always return `"1"`, not `"01"`, so I think `01` is incorrect. Although, `12` for `December` would be correct.

Some investigation would be good. If it's a fix we could include as part of this PR I'd prefer.

hippich: well... 01 and 1 do not give incorrect value, where 2016 and 20 are incorrect values. this is what i was trying to avoid in first place.

I.e. people see format suggestion as `mm/dd/yy` and they might enter `1/12/16`, or  `01/12/16`, or `1/12/2016`. Only last one will produce incorrect Date value as a result of parsing.

rxaviers:
First things first... `strict` parsing has been proposed for solving the issue. I think we all should get aligned on the definition of `strict`. What does it mean here? Does it really solve the issue?

My understanding is that `strict` parsing means the function will return `null` if the provided value isn't strictly what's supposed to be (if it's not strictly what the formatter returns using the same pattern). The opposite of that is `lenient`.

For example, using `strict` should:
- Prevent parsing date `"01/01/2012"` when using pattern `"M/d/yy"`` (wrong length) --- as provided by @hippich in the documentation.
- Prevent parsing date `"01/01/12 foo"` when using pattern `"M/d/yy"`` (extra trailing characters).
- Prevent parsing "01/01" when using pattern `"M/d"` (wrong padding).

The current parser is more strict than lenient in the sense that it will not try to guess for different field lengths (`2016`, `16`, `Jan`, `January`) or for different patterns. Pretty much the only lenient exceptions are padding and trailing characters.

Although Unicode recommends lenient parsing ([link](http://www.unicode.org/reports/tr35/tr35-dates.html#Parsing_Dates_Times)), it says it's "especially complicated, due to the large number of possible fields and formats". Indeed, we would have to use 8 different strict patterns to check for @hippich's example: padded-month/unpadded-month * padded-day/unpadded-day * 2-digits-year/4-digits-year. It grows exponentially.

As far as I understood, @hippich's problem is solved by checking for extra trailing characters only. I think we could solve this as part of the regular parsing, which is already pretty much strict (without the need for the additional `strict` option). The current PR is pretty much a solution against trailing characters if I'm not wrong.

In order to make this change part of the regular parsing, the only question is whether there's benefit to allow extra trailing characters when parsing. I can't think of any.

hippich:
I guess term "strict" chosen by me was not right. What I was trying to avoid is incorrect parsing dates like `12/20/2016` to `Sun Dec 20 2020 00:00:00 GMT-0600 (CST)` when pattern like `M/d/yy` is used.

I guess this is always incorrect result, so probably you are right - may be check for trailing characters should be always applied, instead of adding addition `strict` option, as in my proposed patch.

That would make change much simpler, but people currently using library might be caught unexpected with this change. Interested in your opinion.

The reason why I am against checking zeros padding for month and other parts of the date, for example, is because this will be applied to human-entered content in our case. And users very likely to pad month. I think current parses is on a perfect middle ground between leniency and strictness, except for failing edge-case, where trailing characters are left.

rxaviers:
> I guess this is always incorrect result, so probably you are right - may be check for trailing characters should be always applied, instead of adding addition strict option, as in my proposed patch.

> That would make change much simpler, but people currently using library might be caught unexpected with this change. Interested in your opinion.

@jzaefferer @jquery/globalize any input?

Summarizing the above, currently parser ignores any trailing characters when parsing, so `"2016 foo bar"` is successfully parsed into 2016 considering pattern `y` (year). Do you see any problem on changing the parser to actually return `null` on such (invalid) case? There's a side effect of ignoring trailing characters that lead to incorrectly parsing input (e.g., original error of this PR). Changing this behavior could impact current usage, but I think the change is correct and worth making. This could go in the 1.1.0 release.

> The reason why I am against checking zeros padding for month and other parts of the date, for example, is because this will be applied to human-entered content in our case. And users very likely to pad month. I think current parses is on a perfect middle ground between leniency and strictness, except for failing edge-case, where trailing characters are left.

It makes sense and I'm ok with that.

rxaviers:
Edit: Summarizing the above, currently parser ignores any trailing characters when parsing, so `"2016 foo bar"` is successfully parsed into 2016 considering pattern `y` (year). Do you see any problem on changing the parser to actually return `null` on such (invalid) case? Changing this behavior could impact current usage, but I think the change is correct and worth making. This could go in the 1.1.0 release.

This update fixes a side effect that leads to incorrect parsing, e.g., original error of this PR.

hippich:
Any update on decision? :) Need this functionality in our app and wanted to make sure it will be inline with what globalize.js maintainers wants it to be.

rxaviers:
@hippich let's go with the minimal change (trailing characters = invalid) in regular parsing (no need for the `strict` option). Please, just let me know when you think it's ready for another round of review. Thanks

hippich:
Done :)

rxaviers:
One small thing above. Other than that it LGTM. :clap:

hippich:
Done :)

rxaviers:
Merged, thanks.

Issue #561, opened 2015-12-03
Globalize 1.0.0 does not work the the latest es5-shim


Nexith:
When using ```globalize 1.0.0``` with the latest ```es5-shim``` it throws ```throws new TypeError(ERR_ACCESSORS_NOT_SUPPORTED);``` on line 55 in plural.js.

I first created an issue in the ```es5-shim``` repo, see: es-shims/es5-shim#363 for more information.

ljharb:
Note that this means globalize doesn't support IE 8, which has defineProperty but does not support defining properties on non-DOM elements.

rxaviers:
This regression has been introduced by 7e1cee3b, which updated make-plural from `2.1.2` to `3.0.0`. So, the last Globalize that was compatible with IE8 was `1.0.0-alpha.20`.

rxaviers:
FWIW, this regression should affect the plural module only.

rxaviers:
I plan a patch fix `1.0.1` that will bring back support for IE8. PR #579.

ericelliott:
Looks like that PR was merged. Should this issue be closed?

rxaviers:
Yeap, thanks

rxaviers:
`1.0.1` has been released.

PR #578, opened 2016-01-17
Various: Drop support for IE8

### LINKED ISSUE #571: Drop IE8 support
Currently, IE8 support means two things:
- In the docs, we have instructions for the inclusion of ES5 polyfill in such case;
- In the tests, we have a special trigger for IE8 that uses ES5 polyfill;

We could drop these by dropping support for IE8.

rxaviers:
Ref #561
Fixes #571


rxaviers:
We can't simply remove `test/functional-es5-shim.html` file due to QUnit/phantomjs ([link](https://github.com/jquery/globalize/blob/7ff48e3cc92f539c894ec7fd79dd7b2e0e353ecd/Gruntfile.js#L105-L107)). @jzaefferer, it seems like there's a phantomjs2 ([source](https://github.com/ariya/phantomjs/tree/2.0.0)), any updates on QUnit supporting it? Thanks

Arkni:
Maybe you also need to remove `es5-shim` from `bower.json`.

rxaviers:
> Maybe you also need to remove `es5-shim` from `bower.json`.

That was also my original idea, but can't right now due to https://github.com/jquery/globalize/blob/7ff48e3cc92f539c894ec7fd79dd7b2e0e353ecd/Gruntfile.js#L105-L107.

Arkni:
Just checked their repo and they claim it's fixed in version `2.0`.
So an upgrade may be needed here.

rxaviers:
Yeap, thanks, which is the question I made to @jzaefferer above :smile:

Arkni:
Haha, didn't notice your above comment :sweat_smile:

rxaviers:
:wink:

jzaefferer:
Looks like phantomjs 2.0 is too buggy and 2.1 isn't done yet: https://github.com/gruntjs/grunt-lib-phantomjs/issues/74#issuecomment-148767838 and https://github.com/ariya/phantomjs/issues/12970 (2.0.1 won't happen). 2.1.0 might happen this month: https://github.com/ariya/phantomjs/issues/12970#issuecomment-169539354 

Once that is out, updating grunt-lib-phantomjs seems straight forward enough: https://github.com/gruntjs/grunt-lib-phantomjs/issues/74#issuecomment-156092326

Afterwards grunt-contrib-qunit could be updated accordingly...

rxaviers:
Thanks @jzaefferer

Arkni:
@rxaviers 
Just to let you now, the latest version of `grunt-contrib-qunit` (`v1.0.1`) uses `PhantomJS v2.1.1`.

rxaviers:
Thanks @Arkni

Issue #582, opened 2016-01-28
globalize.js 1.1.0 fails when uglified


hippich:
runtimeBind relies on existence of function name, which is being stripped when source code is uglified. Interested in opinion of maintenaners - should globalize.js adjust its code, or uglify users deal with it via custom settings to avoid this type of code manipulation?

rxaviers:
In the case of Globalize being uglified, I believe we can assume it's already in production, so runtimeBind can be complely ignored. So, I think Globalize should handle that, i.e., runtimeBind should abort in case the function had its name stripped. Does that make sense? Are you interested on submitting a PR? Thanks

hippich:
I need to read code a bit to really understand what runtimeBind does :) will be back

hippich:
I am kinda lost... What runtimeBind, runtimeKey, generatorString, etc are used? So far i find traces of it in tests.. Are these just for tests?

rxaviers:
The main goal of these functions are to allow generating runtime compiled code. More info can be found on #436, on [examples/globalize-compiler](https://github.com/jquery/globalize/tree/master/examples/globalize-compiler), and on https://github.com/jquery-support/globalize-compiler. Please, just let me know if you have any further questions.

hippich:
so.... after looking through it, it appears that my project can't benefit from "compiling", since we implement our own localization framework and invoke globalize dynamically when needed. same applies to formatters - we generate necessary formatter and cache it and then using cached version.

i even tried to run globalize-compiler as is on our code base and got only `Error: No formatters or parsers has been provided` error.

So should in PR you mentioned, when name of the function is missing, just skip the rest of the function code and return passed function right away? I.e. runtimeBind still will be there, but will not do anything.

hippich:
how about that @rxaviers  ?

hippich:
I am gonna make private fork of globalize for now to be able to use it in our project. As soon as we can decide on this issue - I will switch back. Hope to hear from you guys soon.

sergiu-paraschiv:
Why don't you simply disable function name stripping in your minifier of choice until there's a clean workaround? I'm using UflifyJS2 and it has a `--keep-fnames` flag that does just this. In my case the size increase of the minified code was < 5%.

hippich:
Thanks @sergiu-paraschiv, I did not think about that! I will check how much size is affected by keeping function names.

PR #584, opened 2016-01-29
workaround for exception when globalize code is uglified

hippich:
#582


rxaviers:
Thank you @hippich for finding this issue and providing a fix. I've left a couple of comments above, but overall it looks good to me.

rxaviers:
That was quick, thanks

hippich:
:) thanks for spotting it. IE comment wasn't supposed to be deleted :)

rxaviers:
Thank you very much, landed.

PR #586, opened 2016-01-30
Date: Parse February correctly in leap years.

### LINKED ISSUE #585: February Date Parsing
Hello,

I'm running into an issue when parsing dates in Februaries of non-leap years, when the current date is 29-31 of the current month (today is the 29th day of January, for example).  Basically if I choose to parse the date of `2/1/2015` in en with a `date: "short"` format, I get `Sun Mar 01 2015 00:00:00 GMT-0600 (Central Standard Time)`.  This is true for when trying to parse any other day in any other February of non-leap years.  This might not be an issue when we are actually in non-leap year years, as parsing when I set the date variable in `dateParse` to Jan 29, 2015 seems to work with parsing `2/1/2015`.  Parsing `2/1/2016` will work as well.

I'm on the most recent release of globalize, cldrjs, and the cldr-data.  That said, this seems like it was an issue back to at least 1.0 alpha 9.

I plan on investigating this further, but I do appreciate any info or other findings.

Thanks!

navels:
When setting the month, the date parser may set the day to the last day
of the month.  If the year has not yet been set, this will be the last
day of the month in the current year.  So for example, in 2016 "2/1/15"
would be month-parsed to "Feb 29, 2016", then year-parsed to
"Mar 1, 2015".  The fix is set the month after the year has been set.

Also added a test for this case.

Fixes #585.

### DIFF
--- a/test/functional/date/parse-date.js
+++ b/test/functional/date/parse-date.js
@@ -20,7 +20,7 @@ define([
 	enNumbers, enTimeZoneNames, ptCaGregorian, ptNumbers, likelySubtags, numberingSystems, timeData,
 	weekData, util ) {
 
-var ar, date;
+var ar, date, FakeDate;
 
 function extraSetup() {
 	Globalize.load(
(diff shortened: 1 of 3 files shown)


test/functional/date/parse-date.js
```diff
@@ -139,6 +169,27 @@ QUnit.test( "should parse date presets", function( assert ) {
...
+	date = new Date( 2015, 1, 15 );
+	date = startOf( date, "day" );
+	assertParseDate( assert, "Wednesday, February 15, 2015", { date: "full" }, date );
+	assertParseDate( assert, "February 15, 2015", { date: "long" }, date );
+	assertParseDate( assert, "Feb 15, 2015", { date: "medium" }, date );
+	assertParseDate( assert, "2/15/15", { date: "short" }, date );
+
+	Date = OrigDate;
+});
+
```
rxaviers: Please, could you move this test from `functional` to `unit` ([file](https://github.com/jquery/globalize/blob/master/test/unit/date/parse.js#L148))? Unit tests are better to keep the specifics while functional tests are better to test bigger integrations.

rxaviers:
Awesome @navels, thanks for your fix. Great job.

test/functional/date/parse-date.js
navels: Yep, will do.

rxaviers: Thx

navels:
Happy to help, sorry about the period in the commit message :)

rxaviers:
Excellent, thanks for you help! Landed.

PR #595, opened 2016-03-09
Fix broken link in the date-formatter doc

gcastre:
The current link for the CLDR document of the skeleton patterns is actually broken because of a new line at the end of the url in the markdown (Firefox / Chrome / IE11). The url is unicode.o%0arg instead of unicode.org due to the newline being encoded.

https://github.com/jquery/globalize/blob/master/doc/api/date/date-formatter.md#parameters


gcastre:
Can't go to the page for status of CLA check, got a page not found.

jzaefferer:
Looks good to me.

rxaviers:
Landed, thanks.

Issue #539, opened 2015-10-18
Message formatting dependant on base text direction (Bidi support)


ashensis:
Issue of message formatting constitutes quite a challenge from Bidi perspective.
The primary consideration is, in which text direction message is supposed to be displayed.(in DOM model, the HTML element text direction is controlled either by 'dir' attribute or by 'direction' css style) 

For detailed explanation of what base text direction term means please see, for instance, pages 5-9 of the following reference document.
https://docs.google.com/document/d/1dDrSwimrQbpbXybhMYDEiJXLeeqvelnY4cc8HeCP1Nk/edit?usp=sharing

1. Let us consider for illustration the parametrized message:
Globalize.loadMessages({he: {
    error: "THE FILE {0} HASN'T BEEN FOUND",
}});
where upper case stands for Arabic/Hebrew.
Now if file pathe which is going to substitute parameter {0} is in English (like c:\file.txt) and this whole formatted message is supposed to be displayed in left-to-right text direction, the display will be senceless - the reading will be unintelligible.
Globalize("he").messageFormatter("error")("c:\file.txt");

2. In addition to the aforementioned 'use case', there is another typical usage of message formatting (let us call it bread crumb pattern) that may constitute a problem from base text direction prespective, although it may be viewed as a particular sub-case of parametric message.
Let us consider the following message (as is expected to be displayed):
Globalize.loadMessages({he: {
    breadcrumb: "{0} >> {1} >> {2}",
}});

Even if formatted message with substitutions is supposed to be displayed in left-to-right text direction, the outcome will be rather unpredictable depending on substituting text as follows: (upper case stands for Arabic/Hebrew characters, lower for English)

Globalize("he").messageFormatter("breadcrumb")([ "first", "second", "third" ])
will produce on left-to-right display: "first >> second >> third"

Globalize("he").messageFormatter("breadcrumb")([ "FIRST", "SECOND", "THIRD" ])
will produce on left-to-right display: "THIRD << SECOND << FIRST"

Globalize("he").messageFormatter("breadcrumb")([ "first", "SECOND", "THIRD" ])
will produce on left-to-right display: "first >> SECOND << THIRD"
(please note the order of substitutions on display as well as hectic directionality of <<)
In case like 'breadcrumb' the proper segment isolation is required, which may be only achived by using Unicode control characters (aka. UCC)
UCC are to be used in order to cope with the first kind of problems described above. 

The proposal is to augment the 'messageFormatter' (and 'formatMessage') API with
[...]

PR #599, opened 2016-03-22
Globalize:Provide Bidi structured text support to MessageFormat

ashensis:
As per discussion in #539 (comment)
this PR is based on corresponding feature been recently introduced in SlexAxton/messageFormatter.js
(see SlexAxton/messageformat.js#129)
It provides an ability to enforce Bidi structured text in message being formatted by using Unicode Bidi control characters. The upgrading qualities are:

    Prevents intermingling of text contained in placeholders with the rest of message text.
    Maintains the integrity of Bidi text flowing according to directionality of language script. For reference to Bidi structured text see: http://cldr.unicode.org/development/development-process/design-proposals/bidi-handling-of-structured-text

### DIFF
--- a/src/message.js
+++ b/src/message.js
@@ -56,10 +56,12 @@ Globalize.loadMessages = function( json ) {
  *
  * @path [String or Array]
  *
+ * @options [object]
+ *
  * Format a message given its path.
  */
 Globalize.messageFormatter =
-Globalize.prototype.messageFormatter = function( path, isStructuredBidiText ) {
+Globalize.prototype.messageFormatter = function( path, options ) {
 	var cldr, messageFormat, formatter, message, pluralGenerator, returnFn,
 		args = slice.call( arguments, 0 );
 
@@ -88,7 +90,7 @@ Globalize.prototype.messageFormatter = function( path, isStructuredBidiText ) {
 		createErrorPluralModulePresence;
 
 	messageFormat = new MessageFormat( cldr.locale, pluralGenerator );
-	if ( typeof isStructuredBidiText === "boolean" && isStructuredBidiText ) {
+	if ( options && ( options.setBiDiSupport === true ) ) {
 		messageFormat.setBiDiSupport( true );
 	}
 	formatter = messageFormat.compile( message );
@@ -112,7 +114,7 @@ Globalize.prototype.messageFormatter = function( path, isStructuredBidiText ) {
  */
 Globalize.formatMessage =
 Globalize.prototype.formatMessage = function( path /* , variables */ ) {
-	return ( arguments[ 1 ] && typeof arguments[ 1 ] === "boolean" ) ?
+	return ( arguments[ 1 ] && arguments[ 1 ].setBiDiSupport === true ) ?
 		this.messageFormatter( path, arguments[ 1 ] ).apply( {}, slice.call( arguments, 2 ) ) :
 		this.messageFormatter( path ).apply( {}, slice.call( arguments, 1 ) );
 };
--- a/src/message.js
+++ b/src/message.js
@@ -59,8 +59,8 @@ Globalize.loadMessages = function( json ) {
  * Format a message given its path.
  */
 Globalize.messageFormatter =
-Globalize.prototype.messageFormatter = function( path ) {
-	var cldr, formatter, message, pluralGenerator, returnFn,
+Globalize.prototype.messageFormatter = function( path, isStructuredBidiText ) {
+	var cldr, messageFormat, formatter, message, pluralGenerator, returnFn,
 		args = slice.call( arguments, 0 );
 
 	validateParameterPresence( path, "path" );
@@ -87,7 +87,11 @@ Globalize.prototype.messageFormatter = function( path ) {
 		this.pluralGenerator() :
 		createErrorPluralModulePresence;
 
-	formatter = new MessageFormat( cldr.locale, pluralGenerator ).compile( message );
+	messageFormat = new MessageFormat( cldr.locale, pluralGenerator );
+	if ( typeof isStructuredBidiText === "boolean" && isStructuredBidiText ) {
+		messageFormat.setBiDiSupport( true );
+	}
+	formatter = messageFormat.compile( message );
 
 	returnFn = messageFormatterFn( formatter );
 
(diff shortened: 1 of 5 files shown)


test/functional/message/message-formatter.js
```diff
@@ -175,6 +178,13 @@ QUnit.test( "should support ICU message format", function( assert ) {
 	assert.equal( like({ count: 3 }), "You and 2 others liked this" );
 });
 
+QUnit.test( "should support Bidi structured text", function( assert ) {
+	assert.equal(
+		Globalize( "he" ).messageFormatter( "helloArray", true )( "Beethoven", "Mozart" ),
+		"\u05e9\u05dc\u05d5\u05dd, \u200FBeethoven\u200F & \u200FMozart\u200F"
```
rxaviers: Can we replace `\u05e9\u05dc\u05d5\u05dd` with real characters like `שלום` and do the same for all other visual characters?

src/message.js
```diff
@@ -59,8 +59,8 @@ Globalize.loadMessages = function( json ) {
  * Format a message given its path.
  */
 Globalize.messageFormatter =
-Globalize.prototype.messageFormatter = function( path ) {
-	var cldr, formatter, message, pluralGenerator, returnFn,
+Globalize.prototype.messageFormatter = function( path, isStructuredBidiText ) {
```
rxaviers: Instead of the `isStructuredBidiText` argument, let's use an `options` object like `{explicitBiDi: true}`. Feel free to suggest a better (more appropriate) option name.

rxaviers: Another name: `setBiDisupport` (similar to what you used in messageformat).

rxaviers: cc @jzaefferer

src/message.js
```diff
@@ -87,7 +87,11 @@ Globalize.prototype.messageFormatter = function( path ) {
 		this.pluralGenerator() :
 		createErrorPluralModulePresence;
 
-	formatter = new MessageFormat( cldr.locale, pluralGenerator ).compile( message );
+	messageFormat = new MessageFormat( cldr.locale, pluralGenerator );
+	if ( typeof isStructuredBidiText === "boolean" && isStructuredBidiText ) {
```
rxaviers: The condition `isStructuredBidiText === true` should be sufficient.

src/message.js
```diff
@@ -108,7 +112,9 @@ Globalize.prototype.messageFormatter = function( path ) {
  */
 Globalize.formatMessage =
 Globalize.prototype.formatMessage = function( path /* , variables */ ) {
-	return this.messageFormatter( path ).apply( {}, slice.call( arguments, 1 ) );
+	return ( arguments[ 1 ] && typeof arguments[ 1 ] === "boolean" ) ?
+		this.messageFormatter( path, arguments[ 1 ] ).apply( {}, slice.call( arguments, 2 ) ) :
+		this.messageFormatter( path ).apply( {}, slice.call( arguments, 1 ) );
```
rxaviers: I think we can safely use the below instead:

```diff
- return this.messageFormatter( path ).apply( {}, slice.call( arguments, 1 ) );
+ return this.messageFormatter( path ).apply( {}, slice.call( arguments, arguments.length - 1 ) );
```

rxaviers:
@ashensis much better :smile: 

I left a couple of comments, please just let me know if you have any questions.

tomerm:
Leaving a dummy comment to assure future automatic notifications on this PR.

src/message.js
ashensis: I doubt this, consider please the following case where number of formatMessage optional variables is more then one:
Globalize.loadMessages({ en: {breadcrumb: "{0} >> {1} >> {2} >> {3}" ....
Globalize( "en" ).formatMessage( "breadcrumb", "first", "second", "third" , "forth" );
Here arguments.length - 1 === 4 and your proposal will cause incorrect slicing
So probably something like this will be OK?
	return ( arguments[ 1 ]  === true ) ?
		this.messageFormatter( path, arguments[ 1 ] ).apply( {}, slice.call( arguments, 2 ) ) :
		this.messageFormatter( path ).apply( {}, slice.call( arguments, 1 ) );

test/functional/message/message-formatter.js
ashensis: But this involves changing corresponding files to Unicode format, would you mind this?

src/message.js
rxaviers: Nope, this is incorrect. It doesn't matter how big the object/JSON/hash is, it's still one only argument.

test/functional/message/message-formatter.js
rxaviers: It's already `utf-8` to me, so no problem. cc/ @jzaefferer for confirmation.

src/message.js
ashensis: I double checked this with FF Firebug and it proved to be multiple arguments, i.e. in my use case 'arguments' constitutes the plain array object with length corresponding to number of parameters of 'formatMessage' call.
BTW, I probably ought to add this additional parameter explicitly just as it has been done for 'messageFormatter'?

test/functional/message/message-formatter.js
ashensis: Strange, on Windows if one saves some file as having either Unicode or utf-8 encoding using Notepad, it adds special BOM characters at the beginning, and I do not se any in 
message-formatter.js
But, after all, having sample of Hebrew or Arabic text isn't important that much, so I can just use English in this test file.

rxaviers: So please leave it as is. I can do this change later.

src/message.js
rxaviers: Can you share your test with me please?

ashensis: The example is just as I pointed at the beginning of this thread.
You may add in your 
\globalize\examples\plain-javascript\index.html the following:
	Globalize.loadMessages({
		en: {
			breadcrumb: "{0} >> {1} >> {2} >> {3}",
			}
		}		
	});
	var fmOut = Globalize( "en" ).formatMessage( "breadcrumb", "first", "second", "third", "forth"  );

ashensis:
Hi @rxaviers 
I did requested changes, could you please have a look.

rxaviers:
This PR is going to be continued in #623. Closing the latter in favor of the former.

rxaviers:
Thanks for the contribution so far. There are remaining work that needs to be done before we merge this PR into master, which have been addressed in #623 as a continuation of this PR.

PR #604, opened 2016-04-03
DateTimeFormatter: Dynamically augment skeleton

### LINKED ISSUE #271: Date (skeletons): dynamically augment availableFormats

Strate:
Fixes #271 
This is rework of #462.

I have implemented another skeleton pattern comparison algorithm, which fixes a lot of issues I have found in #462 implementation.

Also I have added a lot of tests, which were tested manually across [Intl.DateTimeFormat](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat) and [ICU Flex Test](https://ssl.icu-project.org/icu4jweb/flexTest.jsp).

But not all resulst same from Flex Test and my implementation. This is difference table:

| Skeleton | Flex Pattern | Globalize Pattern | Comment |
| --- | --- | --- | --- |
| `GyMMMEd` | [`EEE, MMM d, y G`](https://ssl.icu-project.org/icu4jweb/flexTest.jsp?pat=GyMMMEd&_=en) | `E, MMM d, y G` | There is explicit available format in cldr-data, which used by globalize |
| `GyMMMEdhms` | [`EEE, MMM d, y G, h:mm:ss a`](https://ssl.icu-project.org/icu4jweb/flexTest.jsp?pat=GyMMMEdhms&_=en) | `E, MMM d, y G, h:mm:ss a` | This is combined `GyMMMEd` and `hms` patterns, which are both available in cldr-data and merged with `dateTimeFormat.medium` |
| `hhmm` | [`h:mm a`](https://ssl.icu-project.org/icu4jweb/flexTest.jsp?pat=hhmm&_=en) | `hh:mm a` | Globalize expand matched `h:mm a` to `hh:mm a` because user requested `hh` in his pattern |
| `EHmms` | [`EEE HH:mm:ss`](https://ssl.icu-project.org/icu4jweb/flexTest.jsp?pat=EHmss&_=en) | `E HH:mm:ss` | There is explicit entry in cldr-data, which matches `EHms` to `E HH:mm:ss`. Because in result pattern `mm` already present, we don't need to augment any more. |

And all other differences are similar. In fact, I am really can not understand, why flexTest augments `E` to `EEE`, if you know why please tell me :-)

### DIFF
--- a/src/date/expand-pattern.js
+++ b/src/date/expand-pattern.js
@@ -1,7 +1,8 @@
 define([
 	"../common/format-message",
-	"../common/create-error/invalid-parameter-value"
-], function( formatMessage, createErrorInvalidParameterValue ) {
+	"../common/create-error/invalid-parameter-value",
+	"./pattern-re"
+], function( formatMessage, createErrorInvalidParameterValue, datePatternRe ) {
 
 /**
  * expandPattern( options, cldr )
@@ -26,7 +27,7 @@ define([
  */
 
 return function( options, cldr ) {
-	var dateSkeleton, result, skeleton, timeSkeleton, type;
+	var dateSkeleton, result, skeleton, timeSkeleton, type, pattern;
 
 	function combineDateTime( type, datePattern, timePattern ) {
 		return formatMessage(
@@ -36,6 +37,89 @@ return function( options, cldr ) {
 			]),
 			[ timePattern, datePattern ]
 		);
+	}
+
+	function getBestMatchPattern( path, skeleton ) {
+		var availableFormats, ratedFormats, format;
+
+		pattern = cldr.main([ path, skeleton ]);
+
+		if (skeleton && pattern === undefined) {
+			availableFormats = cldr.main([ path ]);
+			ratedFormats = [];
+
+			for (format in availableFormats) {
+				ratedFormats.push({
+					format: format,
+					pattern: availableFormats[format],
+					rate: compareFormats(skeleton, format)
+				});
+			}
+
+			ratedFormats.sort(function(a, b) {
+				return a.rate - b.rate;
+			});
+
+			if (ratedFormats.length) {
+				pattern = augmentFormat(skeleton, ratedFormats[0].pattern);
+			}
+		}
+
+		return pattern;
+	}
+
+	function compareFormats( a, b ) {
+		var distance, minLength, i;
+
+		distance = 1;
+		a = a.match(datePatternRe);
+		b = b.match(datePatternRe);
+		minLength = Math.min(a.length, b.length);
+
+		for (i = 0; i < minLength; i++) {
+			if (a[i].charAt(0) === b[i].charAt(0)) {
+				if (a[i].length === b[i].length) {
+					distance *= 0.25;
+				} else {
+					distance *= 0.75;
+				}
+			} else {
+				distance *= 1.25;
+			}
+		}
+
+		if (a.length === b.length) {
+			distance *= 0.5;
+		}
+
+		return distance;
+	}
+
+	function augmentFormat( requestedSkeleton, bestMatchFormat ) {
+		var originalBestMatchFormat, i, t, j, k, l;
+
+		originalBestMatchFormat = bestMatchFormat;
+		requestedSkeleton = requestedSkeleton.match(datePatternRe);
+		bestMatchFormat = bestMatchFormat.match(datePatternRe);
+
+		for (i = 0, l = bestMatchFormat.length; i < l; i++) {
+			t = bestMatchFormat[i].charAt(0);
+
+			for (j = 0, k = requestedSkeleton.length; j < k; j++) {
+				if (t === requestedSkeleton[j].charAt(0)) {
+					bestMatchFormat[i] = requestedSkeleton[j];
+					break;
+				}
+			}
+		}
+
+		bestMatchFormat = bestMatchFormat.join("");
+
+		if (bestMatchFormat === originalBestMatchFormat) {
+			bestMatchFormat = requestedSkeleton.join("");
+		}
+
+		return bestMatchFormat;
 	}
 
 	switch ( true ) {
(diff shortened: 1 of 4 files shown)


rxaviers:
@Strate, fantastic thanks for your contribution. Please, have you used any external reference (e.g., [UTS#35](http://www.unicode.org/reports/tr35/tr35-dates.html)) for implementing your algorithm? If so, could you add internal comments with some references please?

Strate:
@rxaviers My algorithm is close to http://www.unicode.org/reports/tr35/tr35-dates.html#Matching_Skeletons, but seems with several bugs in it. I do not handle rule "Numeric and text fields are given a larger distance from each other. MMM ≈ MM"

I'll work on it but later (next week I suppose)

src/date/expand-pattern.js
```diff
@@ -38,6 +39,128 @@ return function( options, cldr ) {
...
+		return pattern;
+	}
+
+	function repeatStr( str, count ) {
+		var i, result = "";
+		for ( i = 0; i < count; i++ ) {
+			result = result + str;
+		}
+		return result;
+	}
```
rxaviers: Let's move this function into `./src/util/string/repeat-str.js`. Maybe `[./src/util/string/pad.js](https://github.com/jquery/globalize/blob/master/src/util/string/pad.js)` could reuse it.

src/date/expand-pattern.js
```diff
@@ -48,25 +171,38 @@ return function( options, cldr ) {
 			if ( !result ) {
 				timeSkeleton = skeleton.split( /[^hHKkmsSAzZOvVXx]/ ).slice( -1 )[ 0 ];
 				dateSkeleton = skeleton.split( /[^GyYuUrQqMLlwWdDFgEec]/ )[ 0 ];
```
rxaviers: Let's defer this initialization by moving them both inside the `else` below that will actually use them.

src/date/expand-pattern.js
```diff
@@ -48,25 +171,38 @@ return function( options, cldr ) {
...
+					} else {
+						type = "short";
+					}
+
+					if ( dateSkeleton && timeSkeleton ) {
+						result = combineDateTime( type, dateSkeleton, timeSkeleton );
+					} else {
+						result = dateSkeleton || timeSkeleton;
+					}
+				}
```
rxaviers: In general, the whole block above is getting too indented. Let's change it. Currently, we have something like this:

```js
tryDirectMap()
if (!directMap) {
  tryToMatchIt()
  if (!match) {
    tryToMatchItsIndividualParts()
  }
}
```

Let's re-arrange it so we have:


```js
tryDirectMap()
if (directMap) {
  break;
}

tryToMatchIt()
if (match) {
  break;
}

tryToMatchItsIndividualParts()
```

rxaviers: About the `directMap` part of it... It's no longer needed, given that your `getBestMatchPattern` already checks for that.

rxaviers:
Also check/benchmark these ecma-402 polyfil implementations, can we reuse the idea/algo?
- https://github.com/ibm-js/ecma402/blob/afe84f2c0cc02ea4722fb54c3e023476cc49f09a/Intl.js#L510
- https://github.com/andyearnshaw/Intl.js/blob/master/src/12.datetimeformat.js#L447

src/date/expand-pattern.js
```diff
@@ -38,6 +39,128 @@ return function( options, cldr ) {
...
+			availableFormats = cldr.main([ path ]);
+			ratedFormats = [];
+
+			for ( format in availableFormats ) {
+				ratedFormats.push({
+					format: format,
+					pattern: availableFormats[format],
+					rate: compareFormats( skeleton, format )
+				});
+			}
```
rxaviers: This block seems to me a good candidate to be cached, because it's always the same results for the same locale+calendar right? I'm fine adding a `TODO` comment for now.

rxaviers: Ah! Nevermind, just noticed `rate` is a function of the passed `skeleton`.

src/date/expand-pattern.js
```diff
@@ -38,6 +39,128 @@ return function( options, cldr ) {
...
+					normalizePatternType( matchedType ) === normalizePatternType( requestedType ) &&
+					matchedLength < requestedLength
+				) {
+					bestMatchFormat[i] = repeatStr( matchedType, requestedLength );
+				}
+			}
+		}
+
+		return bestMatchFormat.join( "" );
+	}
```
rxaviers: In general about the functions above, can `.match(datePatternRe)` and `normalizePatternType` be made in advance in the caller function so that you don't need to execute these over and over?

Strate: You mean some sort of caching? But this could be a memory issue...

src/date/expand-pattern.js
Strate: Maybe we could use [`String.prototype.repeat`](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/String/repeat) with some of es6-whim?

jawsthegame:
Any movement on this? I could really use this fix.

rxaviers:
It is stalled. Would you like to take it over? I'd be happy if you could address the above comments. You could create a new pull request on top of this one with additional commits. Just let me know if you have any questions. Thanks

PR #623, opened 2016-07-18
Provide Bidi structured text support to MessageFormat (part 2)

### LINKED ISSUE #539: Message formatting dependant on base text direction (Bidi support)

rxaviers:
This is a continuation of PR #599.

Fixes #539 
Ref #599 

Breaking change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -484,7 +484,7 @@ Read more details about locale at [UTS#35 locale][].
 
   [Read more...](doc/api/message/load-messages.md)
 
-- **`.messageFormatter( path ) ➡ function( [variables] )`**
+- **`.messageFormatter( path [, options] ) ➡ function( [variables] )`**
 
   Return a function that formats a message (using ICU message format pattern)
   given its path and a set of variables into a user-readable string. It supports
@@ -500,9 +500,9 @@ Read more details about locale at [UTS#35 locale][].
 
   [Read more...](doc/api/message/message-formatter.md)
 
-- **`.formatMessage( path [, variables ] )`**
+- **`.formatMessage( path [, variables, options ] )`**
 
-  Alias for `.messageFormatter( path )([ variables ])`.
+  Alias for `.messageFormatter( path [, options] )([ variables ])`.
 
 ### Number module
 
--- a/bower.json
+++ b/bower.json
@@ -13,8 +13,8 @@
   },
   "devDependencies": {
     "es5-shim": "3.4.0",
-    "make-plural": "eemeli/make-plural.js#3.0.0",
-    "messageformat": "SlexAxton/messageformat.js#v0.3.0-1",
+    "make-plural": "eemeli/make-plural.js#^3.0.0-rc4",
+    "messageformat": "SlexAxton/messageformat.js#v0.3.0",
     "qunit": "1.18.0",
     "requirejs": "2.1.20",
     "requirejs-plugins": "1.0.2",
--- a/bower.json
+++ b/bower.json
@@ -13,7 +13,7 @@
   },
   "devDependencies": {
     "es5-shim": "3.4.0",
-    "make-plural": "eemeli/make-plural.js#^3.0.0-rc4",
+    "make-plural": "eemeli/make-plural.js#3.0.0",
     "messageformat": "SlexAxton/messageformat.js#v0.3.0",
     "qunit": "1.18.0",
     "requirejs": "2.1.20",
--- a/doc/api/message/message-formatter.md
+++ b/doc/api/message/message-formatter.md
@@ -1,4 +1,4 @@
-## .messageFormatter( path, options ) ➡ function([ variables ])
+## .messageFormatter( path [, options] ) ➡ function([ variables ])
 
 Return a function that formats a message (using ICU message format pattern)
 given its path and a set of variables into a user-readable string. It supports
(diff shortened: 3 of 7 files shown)


rxaviers:
cc @ashensis @tomerm

rxaviers:
The BiDi support itself is entirely implemented in https://github.com/SlexAxton/messageformat.js/pull/128 and https://github.com/SlexAxton/messageformat.js/pull/129. Therefore, this PR is overall quite simple: it's about updating messageformat.js to the corresponding updated version, adding one argument (`options`) to the message formatter methods, and updating the documentation and tests accordingly.

The API changes are shown below:

```diff
The easy API change:
- .messageFormatter( path ) ➡ function( [variables] )
+ .messageFormatter( path [, options] ) ➡ function( [variables] )

The harder API change:
- .formatMessage( path [, variables ] )
+ .formatMessage( path [, variables, options ] )
```

Although the inclusion of the referred `options` argument is somewhat simple, it could represent a challenge for the `formatMessage` API without the following breaking change:

```javascript
formatter = Globalize( "en" ).messageFormatter( "hello" );
 
// Numbered variables using Array.
formatter([ "Wolfgang", "Amadeus", "Mozart" ]);
// > "Hello, Wolfgang Amadeus Mozart"
 
// Numbered variables using function arguments. (NO LONGER SUPPORTED AFTER THIS CHANGE)
formatter( "Wolfgang", "Amadeus", "Mozart" );
// > "Hello, Wolfgang Amadeus Mozart"
```

Please, does anyone (cc @jquery/globalize) have any objection or different suggestion?

jzaefferer:
Breaking compat on `...arguments` seems fine, but should combine that with a major version bump.

As for the implementation, the diff here looks good to me. I'm only missing a test that covers the example from the documentation, something with RTL.

tomerm:
@ashensis is it something we can provide ? I refer to tests. Many thanks in advance.

rxaviers:
If I recall correctly, @ashensis has some issues with his text editor to include that RTL text in the test file, therefore I've included it via 7f42082 fixup.

I need @tomerm or @ashensis to review my update and confirm if the test expected value is correct.

jzaefferer:
I see, though that's not quite what I was looking for. You've got this in the new docs: https://github.com/jquery/globalize/pull/623/files#diff-7735b03e77c86b04c1be49471c334107R231

I'd like to see that in a unit or functional test as well.

ashensis:
1. RTL text in 7f42082 fixup confirmed to be OK.
2. Regarding @jzaefferer remark on lacking uni test that covers the example from the documentation, both message-formatter.js and format-message.js contains such tests.

rxaviers:
> // > "Third << Second << First"

I see. I believe the engine that interprets the BiDi marks will render the string as that, but the returned String itself isn't reverted. Therefore, the doc needs fix.

@tomerm @ashensis could you please confirm?

ashensis:
Indeed , the string itself isn't reverted, the "Third << Second << First" sequence demonstrates how the text will look on display.

rxaviers:

PR #629, opened 2016-08-05
Module names

SteffenMangold:
Module names should not contain path delimiter: "cldr/event", "cldr-event". This makes a lot of trouble with require.

### DIFF
--- a/src/number.js
+++ b/src/number.js
@@ -18,8 +18,8 @@ define([
 	"./number/symbol",
 	"./util/string/pad",
 
-	"cldr/event",
-	"cldr/supplemental"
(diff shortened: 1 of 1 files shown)


rxaviers:
Please, check [this file](https://github.com/jquery/globalize/blob/master/examples/amd-bower/main.js#L7) for an example on how you should configure your AMD path settings.

Please, make additional comments in case it's still a problem for you. Thanks

razbakov:
@rxaviers can you please provide a working example? I am trying to use globalize in vuejs project with webpack. Is there a way without usage of webpack plugin?

Issue #563, opened 2015-12-04
MessageFormat formatters


nkovacs:
The messageformat library supports custom formatter functions. You could register globalize's formatters, so they could be used in messages, e.g. `"Balance: {0, currency}"`, or `"Posted at {0, datetime, long}"`.

It would also be great if there was a way to pass custom formatter functions to MessageFormat.

rxaviers:
Thanks for your message and sorry for the delayed answer.

Please, use variable replacement instead, e.g.: "Balance: {currency}" or "Posted at {date}" and have the variable formatted using the appropriate formatter in your code, e.g.:

```js
Globalize.formatMessage("message", {date: Globalize.formatDate(new Date())});
```

If you find any problem using variable replacement instead or if you have further questions feel free to post additional comments.

PS:

> The messageformat library supports custom formatter functions

... and I was one of the early pushers for such API to be adopted by SlexAxton/messageformat.js (the libraries Globalize uses for mesage formar under the hoods) ([link](https://github.com/SlexAxton/messageformat.js/issues/65)). :smile: (and Alex an Eemeli did a great work updating the library). Having said that, given variable replacement could be used instead with no prejudice in that case, we opt for that.

If you want to update globalize message format to support such feature, feel free to contribute the change and I'd be happy to consider it: (a) send informal messages first to discuss the new API, then send a pull request with the implementation.

nkovacs:
The problem with using a formatter in the variable is that it doesn't allow you to change the format in the message file. It's hard-coded.
This would not only allow the format to be customized for each language, it would also allow changing it without touching the code. E.g. if you have an interface where an admin can change the message files used by your app, this change would allow an admin to customize the format used in a message.

nkovacs:
I made a quick proof of concept. The issues with it are:
* the formatter function only gets a locale string (or array), so it needs to create a new Globalize instance: https://github.com/nkovacs/globalize/commit/bba99b7a82d25d92818c08114132c2657c7a071a#diff-731a3fca6b201d79e2639fe1456b8787R166
* I added a global object to Globalize in core.js to collect the formatters: https://github.com/nkovacs/globalize/commit/bba99b7a82d25d92818c08114132c2657c7a071a#diff-7eb52b366866677666470e019283c8eaR89. This is probably not the most elegant way to do this.
* The compiler compiles the message `'Hello World {now, date, long}'` into this:
```js
Globalize.b955419430 = messageFormatterFn((function(  ) {
[...]

PR #639, opened 2016-10-15
Message: fix selectOrdinal.

nkovacs:
The plural fuction used for messages only supported cardinals,
not ordinals. This meant that selectOrdinal used the cardinal
rules, not the ordinal rules.

If the type option passed to Globalize.pluralGenerator is "both",
the returned function accepts a second parameter. If this
parameter is true, the function will return the ordinal category
instead of the cardinal category.
This is how messageformat.js expects the plural function to work.

This introduces a minor BC break, because supplemental/ordinals.json must be loaded (I've fixed all the examples), although [the documentation does suggest you need both](https://github.com/jquery/globalize#2-cldr-content).

### DIFF
--- a/examples/amd-bower/main.js
+++ b/examples/amd-bower/main.js
@@ -36,6 +36,7 @@ require([
 	"json!cldr-data/supplemental/currencyData.json",
 	"json!cldr-data/supplemental/likelySubtags.json",
 	"json!cldr-data/supplemental/plurals.json",
+	"json!cldr-data/supplemental/ordinals.json",
 	"json!cldr-data/supplemental/timeData.json",
 	"json!cldr-data/supplemental/weekData.json",
 	"json!messages/en.json",
@@ -49,7 +50,7 @@ require([
 	"globalize/relative-time",
 	"globalize/unit"
 ], function( Globalize, enGregorian, enCurrencies, enDateFields, enNumbers, enUnits, currencyData,
-	likelySubtags, pluralsData, timeData, weekData, messages ) {
+	likelySubtags, pluralsData, ordinalsData, timeData, weekData, messages ) {
 
 	var en, like, number;
 
@@ -63,6 +64,7 @@ require([
 		enUnits,
 		likelySubtags,
 		pluralsData,
+		ordinalsData,
 		timeData,
 		weekData
 	);
--- a/examples/app-npm-webpack/package.json
+++ b/examples/app-npm-webpack/package.json
@@ -2,6 +2,7 @@
   "private": true,
   "devDependencies": {
     "cldr-data": ">=25",
+    "globalize": "^1.1.1",
     "globalize-webpack-plugin": "0.3.x",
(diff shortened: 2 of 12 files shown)


rxaviers:
Related to #563

nkovacs:
Regarding your question there:

> About plural requiring cardinal + ordinal data... I want to avoid that. I'm wondering if `{plural, ...` could use a cardinal formatter, and `{selectordinal, ...` could use a ordinal formatter?

Here are two example messages: 

- `This is the {cat, selectordinal, one{#st} two{#nd} few{#rd} other{#th} } category.`
- `This is the {cat, plural, one{#st} two{#nd} few{#rd} other{#th} } category.`

The problem is that `messageFormatterRuntimeBind` sees this:

- `function (d) { return "This is the " + plural(d.cat, 0, pluralFuncs.en, { one: function() { return number(d.cat) + "st";}, two: function() { return number(d.cat) + "nd";}, few: function() { return number(d.cat) + "rd";}, other: function() { return number(d.cat) + "th";} }, `**1**`) + " category."; }`
- `function (d) { return "This is the " + plural(d.cat, 0, pluralFuncs.en, { one: function() { return number(d.cat) + "st";}, two: function() { return number(d.cat) + "nd";}, few: function() { return number(d.cat) + "rd";}, other: function() { return number(d.cat) + "th";} }) + " category."; }`

I've highlighted the 1 in the first message which indicates to the plural function that it should use ordinals.
This would be difficult to find with a regexp; I'd have to parse the javascript properly to make it 100% correct, and that's just overkill.
With the custom compiler in #563, it should be much easier to do this. So, with that in mind, do you think it would be best to abandon this, and implement it with a separate ordinal formatter as part of #563?

dantman:
So is this going to be fixed? I was just writing tests after integrating Globalize based message handling and realized that selectordinal doesn't actually work. Should I have not picked Globalize to handle i18n messages?

rxaviers:
> This introduces a minor BC break, because supplemental/ordinals.json must be loaded (I've fixed all the examples), although the documentation does suggest you need both.

Thanks for bringing this up. It indeed could break existing apps. We could conditionally-enable it only if such CLDR data is loaded, which would prevent to break existing apps, but internal logic would be a little convoluted and error handling in case someone intends to use ordinal but forgot to load it wouldn't be as great. Today I gave this approach a try, but in the end wasn't happy with the outcome. I'm sorry this took so long.

> With the custom compiler in #563, it should be much easier to do this...

Yeap, that's the way we should proceed. I apologize for closing (instead of merging) this PR after so long. Sorry for the frustration.

dantman:
@rxaviers Not an option in my situation. I'm using this in a system where messages and variable values are input by a user. I cannot tweak a variable input to go through a generator.

rxaviers:
Would you be interested in continue @nkovacs's custom compiler change? (or help @nkovacs in case he is still interested pursuing this)

PR #641, opened 2016-10-15
Currency: Fix bugs with CLDR 30 and arabic locales

### LINKED ISSUE #640: Arabic currency changes in cldr 30 cause E_MISSING_CLDR
In cldr 30, the currencyformat for the arabic number system no longer inherits from the latin ([changeset](http://unicode.org/cldr/trac/changeset/12636/trunk/common/main/root.xml)), so the json data is missing currencySpacing and the accounting format for some arabic locales.

For example, these all fail:

```
var arAE = new Globalize( "ar-AE" );
var arSA = new Globalize( "ar-SA" );

console.log(arAE.formatCurrency(1295, "AED"));
console.log(arAE.formatCurrency(-1295, "AED"));
console.log(arAE.formatCurrency(1295, "AED", { style: "accounting" }));
console.log(arAE.formatCurrency(-1295, "AED", { style: "accounting" }));

console.log(arSA.formatCurrency(1295, "SAR"));
console.log(arSA.formatCurrency(-1295, "SAR"));
console.log(arSA.formatCurrency(1295, "SAR", { style: "accounting" }));
console.log(arSA.formatCurrency(-1295, "SAR", { style: "accounting" }));
```

The webpack example also fails when you try to build production files.

ICU4J and ICU for C fall back to a hard-coded default in this case: http://bugs.icu-project.org/trac/browser/icu4j/trunk/main/classes/core/src/com/ibm/icu/impl/CurrencyData.java#L82

For the missing accounting format, ICU4J falls back to the latin accounting format: http://bugs.icu-project.org/trac/browser/icu4j/trunk/main/classes/core/src/com/ibm/icu/text/NumberFormat.java#L1497
(I couldn't figure out what ICU for C does). 

This StackOverflow question has some more information: http://stackoverflow.com/questions/40056378/bug-hunt-cldr-30-json-data-no-longer-has-currencyspacing-information

nkovacs:
- currencySpacing might be missing; use a default taken from ICU4J
- accounting format might be missing; use standard instead

Fixes #640

This is incomplete, because there are a couple open issues:
- What to do when accounting format is not available? ICU4J falls back to latin, this PR falls back to the standard format. I think that makes more sense.
- The default currencySpacing uses `[:letter:]`, which requires including a huge regexp in the code. The cldr data only contains `[:digit:]` and `[:^S:]`, so the default could also use `[:^S:]`, and the letter regexp could be removed.


rxaviers:
Thank you for your contribution. Will review shortly (~days)...

rxaviers:
Thanks for your contribution @nkovacs, but the problem has been fixed in CLDR 30.0.2 data.

PR #646, opened 2016-11-07
Package.json: Loose cldrjs range by using caret range

### LINKED ISSUE #642: Should globalize rely on a hardcoded version of cldrjs?
I am already using cldr.js in my app, and loading CLDR data files. I wanted to add globalize to the mix and assumed that would work great, since it uses the same backend.

Indeed it does work great - with the small kink that I was using a slightly different version of cldr.js. As a result, the data loaded is not recognized by the private copy that globalize.js gets from NPM.

I was thinking that the cldr.js dependency should be made more loose.

rxaviers:
Fixes #642

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -8,7 +8,7 @@
     "bower.json"
   ],
   "dependencies": {
-    "cldrjs": "0.4.4",
+    "cldrjs": "^0.4.4",
     "cldr-data": ">=25"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "url": "http://github.com/jquery/globalize/issues"
   },
   "dependencies": {
-    "cldrjs": "0.4.4"
+    "cldrjs": "^0.4.4"
   },
   "peerDependencies": {
     "cldr-data": ">=25"


jzaefferer:
I still want to see a 1.0.0 release though. Pre-1.0.0 ranges behave very differently.

rxaviers:
> I still want to see a 1.0.0 release though. Pre-1.0.0 ranges behave very differently.

Please, can you detail your concerns? Is it about using a caret instead of any other more explicit range?

jzaefferer:
Check out https://semver.npmjs.com/ with cldrjs and ^0.3.4. It only matches 0.3.4 till ^0.3.11, but not 0.4.x. Which is quite different for caret ranges from 1.0.0 forward.

rxaviers:
> Check out https://semver.npmjs.com/ with cldrjs and ^0.3.4. It only matches 0.3.4 till ^0.3.11, but not 0.4.x. Which is quite different for caret ranges from 1.0.0 forward.

Not what I asked. 😄 It seems like both of us know that, my question is what's your concern about it? Using ^0.4.6 is equivalent to >=0.4.6 && < 0.5.0. Are you concerned with the range itself or with using `^0.4.6` instead of the explicit `>=0.4.6 && < 0.5.0` (or even something else)?

jzaefferer:
If the goal is too loosen it up, then we're gaining very little here. But that shouldn't stop you from landing it anyway...

rxaviers:
Just wanted to understand the concerns. Again, thanks for the valuable inputs.

PR #647, opened 2016-11-07
Package.json: Drop cldr-data peer dependency

### LINKED ISSUE #638: cldr-data as peer dependency?
Does listing cldr-data as peer dependency make sense? We should rething...

By the time it was adopted there was only cldr-data (i.e., there was no unicode official github/npm repo) and npm didn't support flatten dependencies. Defining cldr-data as peer-dependency was a way of having two libraries using CLDR data to reuse the same cldr-data module (e.g., globalize and cldrjs).

One additional goal of defining cldr-data as a peer dependency was to avoid that applications use incompatible CLDR x Globalize versions. In theory, a CLDR schema change could break compatibility with Globalize. Also, new CLDR rules could appear in future releases and Globalize include support for such features, therefore Globalize could update its CLDR compatible range via the peer dependency.

The existing problem is that applications can use different CLDR sources, e.g., the [Unicode modules](https://github.com/unicode-cldr/cldr-core) or local downloaded/processed data. And there's nothing wrong with doing so, because in Globalize we actually empower such choices. We have cldr-data as a convenience, but it's not mandatory. This is basically the existing conflict. In one hand, we rely on cldr-data module to define peer dependency, but in the other hand we have users that don't use it.

rxaviers:
Fixes #638

### DIFF
--- a/.bowerrc
+++ b/.bowerrc
@@ -1,7 +1,6 @@
 {
   "directory": "external",
   "scripts": {
-    "preinstall": "npm install cldr-data-downloader@0.2.x",
     "postinstall": "node ./node_modules/cldr-data-downloader/bin/download.js -i external/cldr-data/index.json -o external/cldr-data/"
   }
 }
--- a/README.md
+++ b/README.md
@@ -183,6 +183,8 @@ Globalize("es").formatDate(new Date());
 // > "27/11/2015"
 ```
 
+Note `cldr-data` is an optional module, read [CLDR content](#cldr-content) section below for more information on how to get CLDR from different sources.
+
 Read the [Locales section](#locales) for more information about supported locales. For AMD, bower and other usage examples, see [Examples section](#examples).
 
 ### Installation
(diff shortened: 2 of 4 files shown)


jzaefferer:
This looks fine to me, still pending my comment from #646. The Getting Started section could be improved, not sure if that's worth it though. A usability test would be nice.

rxaviers:
> The Getting Started section could be improved, not sure if that's worth it though. A usability test would be nice.

Please can you detail these two points? What could be added to/changed in the getting started? What do you consider a usability test, a demo?

jzaefferer:
The Getting Started example doesn't ever say to put the sample code in a JS file and run it with node. Maybe that's obvious enough, maybe it isn't.

Since we don't know, a usability test could show how well the docs are doing, especially for someone just getting started.

rxaviers:
Ok, this is an area that needs improvement with or without dropping cldr-data as peer dependency, so let's handle it separately. Thanks @jzaefferer.

PR #658, opened 2016-12-14
Fix npmignore

rxaviers:
I've done several tests and was surprised by the several failed attempts to get files ignored in published package.

1. Removed the invalid existing `.npmignore`.
1. Merged the various `.gitignore`s (from examples) into `/.gitignore`, but it didn't help. I don't understand why these entries are simply not taken into account by `npm` (tested v3.9.6 from node 5).
1. Used negate expressions in package.json files entry worked, although it goes against recommendation:

> "The consequences are undefined" if you try to negate any of the files entries (that is, "!foo.js"). Please don't. Use .npmignore.

https://github.com/npm/npm/wiki/Files-and-Ignores#details

PS: By the way, I used `tar -tf $(npm pack)` to test a publish, a workaround for a dry run.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .sizecache.json
-bower_components/
-external
-node_modules
+.tmp-globalize-webpack/
+dist/.build/
+external/
+node_modules/
 tmp/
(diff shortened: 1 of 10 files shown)


rxaviers:
out

rxaviers:
Using this for now, we can revisit this in the future.

PR #663, opened 2017-01-01
Fix offset example

vgrigoriu:
I think there should be one case for 1, and one for 2.

### DIFF
--- a/doc/api/message/message-formatter.md
+++ b/doc/api/message/message-formatter.md
@@ -190,8 +190,8 @@ Globalize.loadMessages({
     likeIncludingMe: [
       "{0, plural, offset:1",
       "     =0 {Be the first to like this}",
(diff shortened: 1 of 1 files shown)


rxaviers:
Hi @vgrigoriu, thanks for your PR but the existing example works just fine to me [1]. Please why do you think this change is needed?

1: This is what I get executing the existing example. Do you get a different result?
```
> likeFormatter(0)
'Be the first to like this'
> likeFormatter(1)
'You liked this'
> likeFormatter(2)
'You and someone else liked this'
> likeFormatter(3)
'You and 2 others liked this'
```

vgrigoriu:
Then I guess I'm confused. Why does `likeFormatter(2)` go to the `one` case? I imagined that `one` and `=1` mean the same thing... Is `one` affected by the offset but `=1` not?

PR #676, opened 2017-01-18
Add "use strict" declarations to runtime files

zbirkenbuel:
Runtime script files are now decorated with `"use strict";` statements to keep them in compliance with standard stylecop and eslint style checks.  Globalize code is already strict mode compliant.  This change will not affect browser compatibility since older browsers will simply ignore the statement.  Currently only modifying the built "runtime" files.

strict mode statements are contained within the module loader and factory function scopes so that they do not interfere with other files loaded in a browser.

### DIFF
--- a/src/build/intro-core-runtime.js
+++ b/src/build/intro-core-runtime.js
@@ -15,6 +15,8 @@
  */
 (function( root, factory ) {
 
+	"use strict";
+
 	// UMD returnExports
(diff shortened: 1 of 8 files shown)


jzaefferer:
About half the time the indentation is inconsistent with the following line of code. Otherwise this looks fine to me.

rxaviers:
Merged. Thank you @zbirkenbuel and @jzaefferer for reviewing it.

PR #687, opened 2017-02-09
Date: Olson-timezone-support (real 'z', 'v', 'V')

rajavelmani:
https://github.com/globalizejs/globalize/issues/340

- Supports `z`, `v`, `V` format with timezone names

---

Added by @rxaviers:

TODO:
- [x] Normal Code
- [x] Unit tests
  - [x] test/unit/date/format.js
  - [x] test/unit/date/format-properties.js
- [x] Functional tests
  - [ ] Test the eventual `options.timeZone`
- [ ] Runtime code
- [ ] Documentation
  - [ ] Document the eventual `options.timeZone`
- [ ] Examples
  - [ ] Add the eventual `options.timeZone`

Follow up items:
- Support the historic metazones based on the given date.


src/date/format-properties.js
```diff
@@ -208,6 +209,56 @@ return function( pattern, cldr ) {
...
+					if ( metaZone ) {
+						properties.exemplarCity = cldr.main(
+							"dates/timeZoneNames/zone/" +
+							properties.tzId +
+							"/exemplarCity"
+							);
+						if ( length < 4 ) {
+							properties.standardTzName = cldr.main(
+								"dates/timeZoneNames/metazone/" +
+								metaZone.usesMetazone._mzone +
```
rxaviers: If you still have to access the `.usesMetazone._mzone` structure, on `metaZone` assignment above you can lookup up to that point.

src/date/format-properties.js
```diff
@@ -208,6 +209,56 @@ return function( pattern, cldr ) {
...
+								"/long/standard"
+								);
+							properties.daylightTzName = cldr.main(
+								"dates/timeZoneNames/metazone/" +
+								metaZone.usesMetazone._mzone +
+								"/long/daylight"
+								);
+						}
```
rxaviers:
```diff
diff --git a/src/date/format-properties.js b/src/date/format-properties.js
index 596a18b..825cee3 100644
--- a/src/date/format-properties.js
+++ b/src/date/format-properties.js
@@ -226,29 +226,18 @@ return function( pattern, cldr, tzId ) {
 							properties.tzId +
 							"/exemplarCity"
 							);
-						if ( length < 4 ) {
-							properties.standardTzName = cldr.main(
-								"dates/timeZoneNames/metazone/" +
-								metaZone.usesMetazone._mzone +
-								"/short/standard"
-								);
-							properties.daylightTzName = cldr.main(
-								"dates/timeZoneNames/metazone/" +
-								metaZone.usesMetazone._mzone +
-								"/short/daylight"
-								);
-						} else {
-							properties.standardTzName = cldr.main(
-								"dates/timeZoneNames/metazone/" +
-								metaZone.usesMetazone._mzone +
-								"/long/standard"
-								);
-							properties.daylightTzName = cldr.main(
-								"dates/timeZoneNames/metazone/" +
-								metaZone.usesMetazone._mzone +
-								"/long/daylight"
-								);
-						}
+						properties.standardTzName = cldr.main([
+							"dates/timeZoneNames/metazone",
+							metaZone.usesMetazone._mzone,
+							length < 4 ? "short" : "long",
+							"standard"
+						]);
+						properties.daylightTzName = cldr.main([
+							"dates/timeZoneNames/metazone",
+							metaZone.usesMetazone._mzone,
+							length < 4 ? "short" : "long",
+							"daylight"
+						]);
 						properties.regionFormatStandard = cldr.main(
 							"dates/timeZoneNames/regionFormat-type-standard"
 							);
```

src/date/format-properties.js
```diff
@@ -208,6 +209,56 @@ return function( pattern, cldr ) {
 
 			// Zone
 			case "z":
+
+				// z...zzz: "{shortRegion}", eg. "PST" or "PDT".
+				// zzzz: "{regionName} {Standard Time}" or "{regionName} {Daylight Time}",
+				// eg. "Pacific Standard Time" or "Pacific Daylight Time".
+				if ( properties.tzId ) {
```
rxaviers: We can leave it as is for now, but I have the following comment: who will determine the timezone name is actually date using an imaginary call like `date.getTimezoneName()`. This allows the case where `options.timeZone` is `undefined` that defaults to using local time.

src/date/format-properties.js
```diff
@@ -208,6 +209,56 @@ return function( pattern, cldr ) {
...
+
+					// The latest metazone data of the metazone array.
+					//TODO expand to support the historic metazones based on the given date.
+					var metaZone = cldr.supplemental(
+						[ "metaZones/metazoneInfo/timezone/" + properties.tzId, 0 ]
+						);
+					if ( metaZone ) {
+						properties.exemplarCity = cldr.main(
```
rxaviers: Do we ever need `exemplarCity` if `standardTzName` and `daylightTzName` are found?

src/date/timezone-region-format.js
```diff
@@ -0,0 +1,30 @@
...
+		format;
+	if ( isDST ) {
+		format = ( daylightTzName ) ? daylightTzName :
+		( regionFormatStandard ) ? regionFormatDaylight.replace( /\{0\}/, exemplarCity ) :
+		undefined;
+	} else {
+		format = ( standardTzName ) ? standardTzName :
+		( regionFormatStandard ) ? regionFormatStandard.replace( /\{0\}/, exemplarCity ) :
+		undefined;
+	}
```
rxaviers: It seems to me that at this runtime point we could have done more processing on these properties... I mean, standardTzName (or exemplarCity) could have been incorporated into regionFormatStandard, and daylightTzName (or exemplarCity) could have been incorporated into regionFormatDaylight.

rxaviers: Other points:
- The parentheses on `( standardTzName ) ?` is unnecessary;
- Why the `undefined` case?

test/functional/date/date-formatter.js
```diff
@@ -105,7 +105,8 @@ QUnit.test( "should allow for runtime compilation", function( assert ) {
 					"pm-alt-variant": "pm"
 				},
 				"pattern": "h:mm:ss a",
-				"timeSeparator": ":"
+				"timeSeparator": ":",
+				"tzId": undefined
```
rxaviers: If it's `undefined`, let's simply not include it.

test/unit/date/format.js
```diff
@@ -464,6 +480,26 @@ QUnit.test( "should format timezone (z)", function( assert ) {
...
+	assert.dateFormat( date, "OO", cldr, "GMT-3" );
+	assert.dateFormat( date, "OOO", cldr, "GMT-3" );
+	assert.dateFormat( date, "OOOO", cldr, "GMT-03:00" );
+
+	date = new FakeDate( 11 );
+	assert.dateFormat( date, "O", cldr, "GMT+11" );
+	assert.dateFormat( date, "OO", cldr, "GMT+11" );
+	assert.dateFormat( date, "OOO", cldr, "GMT+11" );
+	assert.dateFormat( date, "OOOO", cldr, "GMT+11:00" );
+});
```
rxaviers: There's already a `"should format timezone (O)"` section above. Instead of duplicating it, let's simply incorporate the missing pieces there please.

src/date/format.js
```diff
@@ -207,6 +208,25 @@ return function( date, numberFormatters, properties ) {
 
 			// Zone
 			case "z":
+
+				// z...zzz: "{shortRegion}", eg. "PST" or "PDT".
+				// zzzz: "{regionName} {Standard Time}" or "{regionName} {Daylight Time}",
+				// eg. "Pacific Standard Time" or "Pacific Daylight Time".
+				if ( properties.tzId ) {
```
rxaviers: Checking for `properties.tzId` isn't valid, because there's a case where `properties.tzId` is truthy and metaZone is unavailable and therefore the properties below are invalid (by looking at your code at date/format-properties.js). I think you should check for something else, e.g., one of the set properties.

src/date/format.js
```diff
@@ -207,6 +208,25 @@ return function( date, numberFormatters, properties ) {
...
+						date,
+						properties.regionFormatStandard,
+						properties.regionFormatDaylight,
+						properties.standardTzName,
+						properties.daylightTzName,
+						properties.exemplarCity
+					);
+					if ( ret ) {
+						break;
+					}
```
rxaviers: Since the properties exist, this should always be truthy.

test/unit/date/format.js
```diff
@@ -52,15 +53,16 @@ Cldr.load(
...
 	timeData,
-	weekData
+	weekData,
+	metaZones
 );
 
 cldr = new Cldr( "en" );
 
-QUnit.assert.dateFormat = function( date, pattern, cldr, expected ) {
+QUnit.assert.dateFormat = function( date, pattern, cldr, expected, tzId ) {
```
rajavelmani: Just to confirm, I added the tzId to the end to avoid the changes to rest of the test cases. if your are ok to change all the test then, i can change all the tests.

src/date/format-properties.js
rajavelmani: just to clarify... do you mean the following?

```diff
+ vat tzId = properties.tzId || date.getTimezone();
-if ( properties.tzId ) 
+if ( tzId )
```

src/date/format-properties.js
rajavelmani: We do not need `exemplarCity` or `regionFormatStandard` and `regionFormatDaylight`. Let me change the code to get these data when `standardTzName` and `daylightTzName` are not found.

src/date.js
```diff
@@ -51,7 +51,7 @@ function validateRequiredCldr( path, value ) {
  */
 Globalize.dateFormatter =
 Globalize.prototype.dateFormatter = function( options ) {
-	var args, cldr, numberFormatters, pad, pattern, properties, returnFn;
+	var args, cldr, numberFormatters, pad, pattern, properties, returnFn, tzId;
```
rajavelmani: i can use `timeZone` as `timezoneName` is locale dependent

test/unit/date/format.js
rajavelmani: I have changed the order of the signature as you mentioned.

test/unit/date/format.js
```diff
@@ -443,6 +445,20 @@ QUnit.test( "should format various milliseconds (A+)", function( assert ) {
...
+QUnit.test( "should format timezone (z)", function( assert ) {
+	var date = new Date();
+	assert.dateFormat(date, "z", cldr, "PST", "America/Los_Angeles");
+	assert.dateFormat(date, "zz", cldr, "PST", "America/Los_Angeles");
+	assert.dateFormat(date, "zzz", cldr, "PST", "America/Los_Angeles");
+	assert.dateFormat(date, "zzzz", cldr, "Pacific Standard Time", "America/Los_Angeles");
+
+	//fall through to 'O' format
+	date = new FakeDate( 0 );
+	assert.dateFormat(date, "z", cldr, "GMT", "Etc/GMT");
```
rajavelmani: i used `"America/Argentina/Buenos_Aires"` which is not supported in CLDR json yet

src/date/timezone-region-format.js
rajavelmani: >It seems to me that at this runtime point we could have done more processing on these properties... I mean, standardTzName (or exemplarCity) could have been incorporated into regionFormatStandard, and daylightTzName (or exemplarCity) could have been incorporated into regionFormatDaylight.

I have optimized the `properties` please take a look and let me know your thoughts

src/date/format.js
rajavelmani: ok let me change this to check `properties.standardTzName` and `properties.daylightTzName`

src/date/format.js
rajavelmani: Yes, this will not be required with check  for `properties.standardTzName and properties.daylightTzName

src/date/format-properties.js
```diff
@@ -208,6 +209,58 @@ return function( pattern, cldr ) {
...
+							length < 4 ? "short" : "long",
+							"standard"
+						]);
+						daylightTzName = cldr.main([
+							"dates/timeZoneNames/metazone",
+							metaZone,
+							length < 4 ? "short" : "long",
+							"daylight"
+						]);
+						if ( !standardTzName && !daylightTzName ) {
```
rxaviers: What should happen for `!standardTzName || !daylightTzName`? It won't fall either in here or in the next `if`. Should we use `examplarCity` if `!standardTzName || !daylightTzName`. Probably, this would only happen in a corrupted CLDR data where there is one content, but not the other.

src/date/format-properties.js
rxaviers: Hi Mani,

I actually mean this:

```diff
if (properties.tzId) {
  date.setTimezone(properties.tzId);
}
...
+ vat tzId = date.getTimezone();
-if ( properties.tzId ) 
+if ( tzId )
```

test/unit/date/format.js
rxaviers: Nice! Though, I'm wondering... Should we use an invalid stuff that will never be supported like `"Foo/Bar"`?

src/date/format-properties.js
```diff
@@ -208,6 +209,58 @@ return function( pattern, cldr ) {
...
+							var exemplarCity = cldr.main(
+								"dates/timeZoneNames/zone/",
```
rxaviers: remove trailing /

src/date/format-properties.js
```diff
@@ -208,6 +209,58 @@ return function( pattern, cldr ) {
...
+								properties.timeZone,
+								"/exemplarCity"
+							);
+							standardTzName = cldr.main(
+								"dates/timeZoneNames/regionFormat-type-standard"
+							)
+							.replace( /\{0\}/, exemplarCity );
```
rxaviers: Instead of replacing `{0}` yourself, you can use [src/common/format-message](https://github.com/globalizejs/globalize/blob/master/src/common/format-message.js) ([example](https://github.com/globalizejs/globalize/blob/master/src/date/expand-pattern.js#L32-L37)).

rxaviers: FWIW, it's an internal utility to help in cases like that. Note, it's not ICU message format :)

src/date/format-properties.js
```diff
@@ -19,11 +19,12 @@ define([
...
  * TODO Support other calendar types.
  */
-return function( pattern, cldr ) {
+return function( pattern, cldr, timeZone ) {
 	var properties = {
 			numberFormatters: {},
 			pattern: pattern,
-			timeSeparator: numberSymbol( "timeSeparator", cldr )
+			timeSeparator: numberSymbol( "timeSeparator", cldr ),
+			timeZone: timeZone
```
rxaviers: Let's skip adding `timeZone` to the properties. We can add it later if required.

test/unit/date/format.js
```diff
@@ -77,517 +80,533 @@ QUnit.module( "Date Format" );
...
  *  Zone
  */
-
 QUnit.test( "should format timezone (z)", function( assert ) {
+	var date = new Date();
+	assert.dateFormat(date, "z", "America/Los_Angeles", cldr, "PST");
+	assert.dateFormat(date, "zz", "America/Los_Angeles", cldr, "PST");
+	assert.dateFormat(date, "zzz", "America/Los_Angeles", cldr, "PST");
+	assert.dateFormat(date, "zzzz", "America/Los_Angeles", cldr, "Pacific Standard Time");
+
```
rxaviers: We need to test the "exemplarCity" case.

src/date/format-properties.js
rajavelmani: If we don't add `timeZone` from `option.timeZone` to `properties.timeZone` we may not be able to do the tests. Can you clarify why we need to skip?

src/date/format-properties.js
rajavelmani: this will help on the scenario like https://github.com/globalizejs/globalize/pull/687/files#diff-df6075946ae91cbb093f63c7597cd746R460 where the timezone id is valid not supported in CLDR.

src/date/format-properties.js
rajavelmani: i just followed from the [format.js](https://github.com/globalizejs/globalize/pull/687/files#diff-3372b7628d9709f40d171d30a27339feR236). but, i can use the format-message as you suggested. Do i change the format.js code too?

src/date/format-properties.js
rxaviers: Note I'm referring to using `||` instead of `&&` in here, so we also catch the additional case where part of the data exists.

test/unit/date/format-properties.js
```diff
@@ -117,4 +119,58 @@ QUnit.test( "should return dayPeriods property for period (a)", function( assert
...
+			length = pattern.length;
+		assert.ok( "genericTzName" in properties( pattern, cldr, timeZone ) );
+
+		if ( length === 1 ) {
+			assert.equal( "PT", properties( pattern, cldr, timeZone ).genericTzName );
+		} else if ( length === 4 ) {
+			assert.equal( "Pacific Time", properties( pattern, cldr, timeZone ).genericTzName );
+		}
+	});
+});
```
rxaviers: Let's add a test to assert `vv` and `vvv` throw.

test/unit/date/format-properties.js
```diff
@@ -117,4 +119,58 @@ QUnit.test( "should return dayPeriods property for period (a)", function( assert
...
+		if ( length === 2 ) {
+			assert.equal( "America/Los_Angeles", properties( pattern, cldr, timeZone ).TzName );
+		} else if ( length === 3 ) {
+			assert.equal( "Los Angeles", properties( pattern, cldr, timeZone ).TzName );
+		} else if ( length === 4 ) {
+			assert.equal( "Los Angeles Time", properties( pattern, cldr, timeZone ).TzName );
+		}
+	});
+});
+
```
rxaviers: Let's add a test to assert `V` throws.

src/date/format-properties.js
```diff
@@ -208,6 +283,63 @@ return function( pattern, cldr ) {
 
 			// Zone
 			case "z":
+				properties.standardTzName = standardTzName;
+				properties.daylightTzName = daylightTzName;
+				break;
```
rxaviers: nitpick: let's add an empty line after `break` and before the next `case`.

src/date/format-properties.js
```diff
@@ -208,6 +283,63 @@ return function( pattern, cldr ) {
...
+						var unKnownExemplarCity = cldr.main([
+							"dates/timeZoneNames/zone/Etc/Unknown/exemplarCity"
+						]);
+						TzName = formatMessage(
+							cldr.main(
+								"dates/timeZoneNames/regionFormat"
+							),
+							[ unKnownExemplarCity ]
+						);
+					}
```
rxaviers: Can we simplify the the above as:

```diff
 					if ( length === 2 ) {
 						TzName = timeZone;
-					}
-
-					if ( exemplarCity ) {
-						if ( length === 3 ) {
-							TzName = exemplarCity;
-						} else if ( length === 4 ) {
-							TzName = formatMessage(
-								cldr.main(
-									"dates/timeZoneNames/regionFormat"
-								),
-								[ exemplarCity ]
-							);
+					} else if ( length === 3 ) {
+						TzName = exemplarCity;
+					} else if ( length === 4 ) {
+						if ( !exemplarCity ) {
+							exemplarCity = cldr.main([
+								"dates/timeZoneNames/zone/Etc/Unknown/exemplarCity"
+							]);
 						}
-					} else if ( !exemplarCity && length === 4 ) {
-						var unKnownExemplarCity = cldr.main([
-							"dates/timeZoneNames/zone/Etc/Unknown/exemplarCity"
-						]);
 						TzName = formatMessage(
 							cldr.main(
 								"dates/timeZoneNames/regionFormat"
 							),
-							[ unKnownExemplarCity ]
+							[ exemplarCity ]
 						);
 					}
 					if ( TzName ) {
```

rxaviers: https://gist.github.com/d6ca0c74e17c203da094a1167b875b9b

test/unit/date/format.js
```diff
@@ -443,25 +473,100 @@ QUnit.test( "should format various milliseconds (A+)", function( assert ) {
...
+	date = new FakeDate( 0 );
+	assert.dateFormatWithTimezone( date, "v", "America/Argentina/Buenos_Aires", cldr, "Unknown City Time" );
+	assert.dateFormatWithTimezone( date, "vvvv", "America/Argentina/Buenos_Aires", cldr, "Unknown City Time" );
+});
+
+QUnit.test( "should format timezone (V)", function( assert ) {
+	var date = new FakeDate( 0 );
+	assert.dateFormatWithTimezone( date, "VV", "America/Los_Angeles", cldr, "America/Los_Angeles" );
+	assert.dateFormatWithTimezone( date, "VVV", "America/Los_Angeles", cldr, "Los Angeles" );
+	assert.dateFormatWithTimezone( date, "VVVV", "America/Los_Angeles", cldr, "Los Angeles Time" );
```
rxaviers: Can we include `VVV` and `VVVV` tests that use unkown city? It could re-use the above America/Argentina/Buenos_Aires, but it's good to have it explicitly tested here (in addition to the above fallback-test). Thanks

test/unit/date/format-properties.js
rxaviers: In here, or where it's possible to test this...

src/date/format-properties.js
```diff
@@ -19,13 +20,28 @@ define([
...
+		exemplarCity;
+
+		if ( timeZone ) {
+
+			// The latest metazone data of the metazone array.
+			//TODO expand to support the historic metazones based on the given date.
+			metaZone = cldr.supplemental([
+				"metaZones/metazoneInfo/timezone", timeZone, 0,
+				"usesMetazone/_mzone"
+			]),
```
rajavelmani: ok let me move this condition into https://github.com/globalizejs/globalize/pull/687/files#diff-a8a17cd910e21f0b9d3e8ffe4fcea2a8R50 to access `chr`

src/date.js
```diff
@@ -35,6 +35,24 @@ function validateRequiredCldr( path, value ) {
 }
 
 /**
+ * .loadIANA( json )
+ *
+ * @json [JSON]
+ *
+ * Load IANA data.
+ */
+Globalize.loadIANA = function( json ) {
```
rxaviers: Pros:
- Letting users to load this data similarly to the way they do with CLDR is good. Therefore we don't embed such in the library.

Cons:
- Differently from CLDR, IANA doesn't provide with an official JSON binding. The one used here is non-official, it is using the JSON bindings from moment-timezone, and could be subject to changes. Ways to mitigate risks here could be 1) to get moment-timezone data put into a separate iana-json library; 2) check if IANA is interested in making it official? Ideas?

src/date.js
```diff
@@ -62,9 +80,12 @@ Globalize.prototype.dateFormatter = function( options ) {
 
 	validateDefaultLocale( cldr );
 
+	timeZone = options.timeZone;
+	validateParameterTypeString( options.timeZone, "options.timeZone" );
```
rxaviers: Now that we the additional option `options.timeZone`, we need to update the above `options = options || { skeleton: "yMd" }`. For example, `dateFormatter({timeZone: "something"})` should also set the default `skeleton: "yMd"`. This default doesn't need to be set if either one is used: skeleton, date, or datetime.

src/date/format-properties.js
```diff
@@ -3,9 +3,10 @@ define([
 	"./pattern-re",
 	"../common/create-error/unsupported-feature",
 	"../number/symbol",
-	"../util/string/pad"
+	"../util/string/pad",
+	"../common/format-message"
```
rxaviers: nitpick: sort these filepaths alphabetically please, therefore `common/format-message` should be included after the `common/..` one above.

test/unit/date/format.js
```diff
@@ -52,15 +53,620 @@ Cldr.load(
...
+	}
+});
+
+// needed for magicDate
+Cldr.load({
+	"globalize-iana": {
+		"zones": [
+			{
+				"name": "America/Los_Angeles",
+				"abbrs": [
```
rxaviers: Should we remove this?

test/unit/date/format.js
```diff
@@ -52,15 +53,620 @@ Cldr.load(
...
+				420,
+				480,
+				420,
+				480,
+				420,
+				480,
+				420,
+				480
+				],
+				"population": 15058000
```
rxaviers: Should we remove this?

test/unit/date/format.js
```diff
@@ -443,25 +1049,111 @@ QUnit.test( "should format various milliseconds (A+)", function( assert ) {
...
+	//which is deprecated in CLDR json and thus no metaZones data
+	date = new FakeDate( 0 );
+	assert.dateFormatWithTimezone( date, "z", "America/Argentina/Buenos_Aires", cldr, "GMT" );
+	assert.dateFormatWithTimezone( date, "zz", "America/Argentina/Buenos_Aires", cldr, "GMT" );
+	assert.dateFormatWithTimezone( date, "zzz", "America/Argentina/Buenos_Aires", cldr, "GMT" );
+	assert.dateFormatWithTimezone( date, "zzzz", "America/Argentina/Buenos_Aires", cldr, "GMT" );
+});
+
+QUnit.test( "should format timezone (v)", function( assert ) {
+	try {
```
rxaviers: It seems like this was used for debug and was left?

rxaviers:
Closed in favor of #701

PR #688, opened 2017-02-12
 type attribute not required in HTML5 

uzerus:
Cos javascript is default HTML scripting language now


rxaviers:
This PR touches 2011 code that doesn't exist in master since a long time. It's outdated. Therefore, it cannot be accepted.

That's the second time you submit this same PR. Please, could you explain your intentions?

Having said that, your contribution on our recent master is welcome. Feel free to send changes for anything you find wrong. Thanks

Issue #292, opened 2014-08-22
Parse: Loose Matching


rxaviers:
Handle [Loose Matching](http://unicode.org/reports/tr35/#Loose_Matching) in all parsing functions, which includes normalizing no-break space to space, different types of dashes, comma used as abbreviation, etc.

References:
- ES6 `String.prototype.normalize` and its polyill [unorm.js](https://github.com/walling/unorm/blob/master/src/unorm.js).
- List of Cf characters: http://unicode.org/cldr/utility/list-unicodeset.jsp?a=%5B%3ACf%3A%5D&g=
- List of Dash characters: http://unicode.org/cldr/utility/list-unicodeset.jsp?a=%5B%3ADash%3A%5D&g=
- Unicode Character Database (UCD) http://www.unicode.org/reports/tr44/
- Unicode Categories Regular Expressions http://www.unicode.org/reports/tr18/#Categories

This might also be useful: http://userguide.icu-project.org/strings/unicodeset and http://userguide.icu-project.org/strings/properties

TODO:
- The problem is that unorm.js currently embeds the normalization lookup data, making it 36.6KB big (minified+gzipped), which is 10x bigger than Globalize and its number module together. While it's not a problem for backend application, it may be way too much for frontend. Out of curiosity, stripping the embedded data out from unorm.js makes it 2.0KB (minified+gzipped).

PS: original issue #288

PR #691, opened 2017-02-20
Number: Fix parser to handle empty string

### LINKED ISSUE #682: New v1.2.* returns 0 when number-parsing empty strings
See 

http://stackoverflow.com/questions/41748476/globalize-bug-in-new-v1-2-parsing-empty-string-returns-0

In short, there seems to be a regression in v1.2.* in that empty strings are now number parsed to 0 instead of NaN This seems rather non-intuitive, and I can't imagine it's intentional.

rxaviers:
Fixes #682
Ref #292
Ref #353

### DIFF
--- a/src/date/tokenizer.js
+++ b/src/date/tokenizer.js
@@ -374,7 +374,7 @@ return function( value, numberParser, properties ) {
 
 			default:
 				token.type = "literal";
-				tokenRe = /./;
+				tokenRe = new RegExp( regexpEscape( current ) );
 		}
 
 		if ( !tokenRe ) {
--- a/src/number/parse-properties.js
+++ b/src/number/parse-properties.js
@@ -135,10 +135,14 @@ return function( pattern, cldr, options ) {
 	if ( !isNaN( minimumSignificantDigits * maximumSignificantDigits ) || /* 1 */
 				maximumFractionDigits /* 2 */ ) {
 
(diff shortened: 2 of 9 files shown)


jzaefferer:
That's a lot of regular expressions that are pretty hard to read. I can't tell where in the tokenizer update you're now handling the empty string. But it seems to be working, so, up to you to add more comments or not.

rxaviers:
Yeap, the regexps are starting to go crazy. The explanation I have is:
- https://github.com/rxaviers/globalize/blob/b1745f1babf7fb9d5a7bcb8293405c3682b0523c/src/number/parse-properties.js#L101 which corresponds to https://github.com/rxaviers/globalize/blob/b1745f1babf7fb9d5a7bcb8293405c3682b0523c/src/number/parse-properties.js#L138-L139
- The extra part has its own comment https://github.com/rxaviers/globalize/blob/b1745f1babf7fb9d5a7bcb8293405c3682b0523c/src/number/parse-properties.js#L141-L143

Any suggestion how to improve that please?

jzaefferer:
I've looked at those, but barely understand anything. I think I'd have to spend a few hours to understand the context, probably not quite worth it as this point.

So, good to go :-)

rxaviers:
Sure thanks for the review

PR #692, opened 2017-02-21
Date: Parse literals, e.g., separators

### LINKED ISSUE #683: Date parser: Should not ignore separators
Parsing a date with correct separator works and is correct:

```js
> Globalize('en').parseDate('1/2/17', {skeleton: 'yMd'})
0017-01-02T02:00:00.000Z
```

Parsing a date with incorrect separator currently works, but it's correct:

```js
> Globalize('en').parseDate('1-2-17', {skeleton: 'yMd'})
0017-01-02T02:00:00.000Z
> Globalize('en').parseDate('1.2.17', {skeleton: 'yMd'})
0017-01-02T02:00:00.000Z
> Globalize('en').parseDate('1.2.2017', {skeleton: 'yMd'})
2017-01-02T02:00:00.000Z
> Globalize('en').parseDate('10.2.2017', {skeleton: 'yMd'})
2017-10-02T03:00:00.000Z
> Globalize('en').parseDate('10a2a2017', {skeleton: 'yMd'})
2017-10-02T03:00:00.000Z
```

rxaviers:
Fixes #683

### DIFF
--- a/test/functional/date/parse-date.js
+++ b/test/functional/date/parse-date.js
@@ -122,10 +122,10 @@ QUnit.test( "should parse time presets", function( assert ) {
 	date.setSeconds( 7 );
 	date = startOf( date, "second" );
 	assertParseDate( assert, "5:35:07 PM", { time: "medium" }, date );
-	assertParseDate( assert, "٥،٣٥،٠٧ م", { time: "medium" }, date, ar );
+	assertParseDate( assert, "٥:٣٥:٠٧ م", { time: "medium" }, date, ar );
 	date = startOf( date, "minute" );
 	assertParseDate( assert, "5:35 PM", { time: "short" }, date );
-	assertParseDate( assert, "٥،٣٥ م", { time: "short" }, date, ar );
+	assertParseDate( assert, "٥:٣٥ م", { time: "short" }, date, ar );
 });
 
 QUnit.test( "should parse date presets", function( assert ) {
--- a/src/date/tokenizer.js
+++ b/src/date/tokenizer.js
@@ -374,7 +374,7 @@ return function( value, numberParser, properties ) {
 
 			default:
 				token.type = "literal";
-				tokenRe = /./;
+				tokenRe = new RegExp( regexpEscape( current ) );
 		}
 
 		if ( !tokenRe ) {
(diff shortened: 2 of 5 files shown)


jzaefferer:
I looked at the diff and didn't find anything fishy. But I don't understand most of what's being changed there, so not sure how valuable that kind of review really is.

rxaviers:
Thanks for checking it out @jzaefferer, the goal is to fix #683 and the relevant change is this one https://github.com/globalizejs/globalize/pull/692/files#diff-4b6bbfed329023ed0ac60539e2ee7e4eR377

test/functional/date/parse-date.js
```diff
@@ -122,10 +122,10 @@ QUnit.test( "should parse time presets", function( assert ) {
 	date.setSeconds( 7 );
 	date = startOf( date, "second" );
 	assertParseDate( assert, "5:35:07 PM", { time: "medium" }, date );
-	assertParseDate( assert, "٥،٣٥،٠٧ م", { time: "medium" }, date, ar );
+	assertParseDate( assert, "٥:٣٥:٠٧ م", { time: "medium" }, date, ar );
```
jzaefferer: Can you explain what this assertion is doing and what you changed here and why? That might help understand the bigger picture.

rxaviers: In previous CLDR (when this test was created), `،` was used as time separator. In recent CLDR, it was reverted to use `:` again. Although, the parser was buggy and didn't actually check for that. It was still passing because whatever character was working there. After the fix, this needed fix.

jzaefferer: Gotcha, thanks. Now this change makes a lot more sense.

jzaefferer:
Besides the "should parse invalid literal as null" test, there doesn't seem to be any new test to support "parse literals" here. Maybe there should be a new functional test for that?

rxaviers:
> Besides the "should parse invalid literal as null" test, there doesn't seem to be any new test to support "parse literals" here.

I've added an additional unit test a4fd84f. Now I ran out of ideas how to better test it. Thanks.

> Maybe there should be a new functional test for that?

In the spirit of not repeating (unnecessary) tests (which I mostly learned from you 😝), I don't see how functional tests can help in this specific fix, but I'm open for ideas (in such case we can add it in a separate PR).

jzaefferer:
Having one extra test seems fine. I don't understand what its doing though.

PR #699, opened 2017-03-02
Date: Filter am/pm data for "a" date pattern

### LINKED ISSUE #509: parseDate returns AM datetime in zh and zh-Hant despite Chinese PM characters
When evaluating amPm, only "pm" is used to determine if the datetime is in the PM. Because "afternoon" exists in zh and zh-Hant, and the Chinese characters for it are exactly the same as those used for "PM", and it alphabetically comes first, it sets the amPm variable to "afternoon" rather than "pm" and therefore incorrectly does not add 12 hours.

rxaviers:
CLDR introduced additional dayPeriods (e.g., morning1, afternoon1,
evening1, etc) along with am and pm. The new dayPeriods should be handle
by "b" and "B" date patterns, while the "a" pattern should still handle
am/pm.

For Chinese, the corresponding value for either pm or afternoon1 is
exactly the same, therefore it causes problem on the reverse lookup when
parsing.

This change fixes the above parsing issue by filtering am/pm only in the
parser properties. It also filters it out on formatting as a simple
optimization (to avoid unnecessary properties).

Fixes #509
Closes #508

### DIFF
--- a/test/unit/date/parse.js
+++ b/test/unit/date/parse.js
@@ -7,6 +7,8 @@ define([
 	"src/date/tokenizer-properties",
 	"json!cldr-data/main/en/ca-gregorian.json",
 	"json!cldr-data/main/en/numbers.json",
+	"json!cldr-data/main/zh/ca-gregorian.json",
+	"json!cldr-data/main/zh/numbers.json",
 	"json!cldr-data/supplemental/likelySubtags.json",
 	"json!cldr-data/supplemental/timeData.json",
 	"json!cldr-data/supplemental/weekData.json",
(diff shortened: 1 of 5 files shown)


jzaefferer:
I looked at the diff. Can't say that I understand much, but also didn't spot anything fishy. Maybe that helps?

test/unit/date/parse.js
```diff
@@ -269,6 +274,8 @@ QUnit.test( "should parse period (a)", function( assert ) {
 	date2 = startOf( date2, "hour" );
 	assertParse( assert, "5 AM", "h a", cldr, date1 );
 	assertParse( assert, "5 PM", "h a", cldr, date2 );
+	assertParse( assert, "上午5", "ah", zh, date1 );
+	assertParse( assert, "下午5", "ah", zh, date2 );
```
jzaefferer: Can you explain what this assertion does? What's the purpose of `assertParse`, what are the arguments doing?

rxaviers: I wish this was something like `expect( parse( "下午5", "ah", zh ) ).to.equalTime( date2 )`. Back to our existing code, `assertParse` is a utility function defined in this same file. It asserts that parsing `"下午"` string in the `"ah"` raw-pattern using the `zh` (Chinese) data equals `date2`.

Issue #406, opened 2015-03-02
Support for short and long decimal formatting


KL-7:
This feature would allow abbreviating big numbers such as `1,234,567` as `1M` or `1 million`.

Short and long decimal formatting is briefly described in "Short Numbers" section of ["Number Patterns"](http://cldr.unicode.org/translation/number-patterns) and the data for it is available in [numbers.json](http://www.unicode.org/repos/cldr-aux/json/25/main/en/numbers.json).

Both formats are implemented in [twitter-cldr-js](https://github.com/twitter/twitter-cldr-js) as subclasses of [AbbreviatedNumberFormatter](https://github.com/twitter/twitter-cldr-js/blob/master/lib/twitter_cldr/js/mustache/numbers/numbers.coffee#L135-L171) if you want to check it out for reference. And the specs are over [here](https://github.com/twitter/twitter-cldr-js/tree/master/spec/js/numbers/abbreviated).

---

Edited by rxaviers:
- Related Ecma proposal https://github.com/tc39/ecma402/issues/37

rxaviers:
:+1:

We need to brainstorm the API. Something to consider: currencies, e.g., `$1.5M`?

M-Areeb:
can i fix this bug

rxaviers:
@M-Areeb feel free to submit a PR with a fix.

rxaviers:
Related Ecma proposal https://github.com/tc39/ecma402/issues/37

necolas:
Hey, please could you give me a few pointers as to how we could get started adding this feature?

rxaviers:
Hi @necolas!

It's great to hear from you. The first step is to define an API. Are we going to reuse number formatter? How can currency and unit formatters reuse this feature? It would be ideal to do a good research on the existing APIs&proposals elsewhere and to get it digested in here, so that we can brainstorm about it. The second step, the implementation, will depend on what we define, so as soon as we're happy with the API I can provide additional implementation pointers. Did I answer your question? Just let me know if you are looking for something else.

necolas:
I think the ECMA proposal APIs looks good enough

jzaefferer:
You mean this one, right?
```js
new Intl.NumberFormat('en', { compact: 'best-fit' }).format(1234000); // 1.2M
new Intl.NumberFormat('en', { compact: '4-digits' }).format(1234000); // 1234K
```
[...]

PR #715, opened 2017-04-27
Numbers: support compact mode

sieverk:
WIP addressing #406 that I wanted to solicit some feedback on. I'm trying to add support for short/long compact numbers (long/short patterns [already included](http://www.unicode.org/repos/cldr-aux/json/25/main/en/numbers.json) in decimalFormats in CLDR). My proposed API is the following:

```
.numberFormatter({ compact: "short" })( 14000000 )
// > "14M"

.numberFormatter({ compact: "long" })( 14000000 )
// > "14 million"
```

The `compact` option plays nicely with the other options with one small asterisk: the `long` pattern can look a bit goofy with percents. ("14 million%") And currency/unit should be able to make use of this options in their number formatters as well.

Because compact patterns can only be determined when the number is in hand, we have to parse/apply that pattern at format-time. And since we can't determine which plural form applies until we have the rounded, decimal-shifted number, we have to check the pattern twice (the first time, figure out the decimal-shift divisor based on the "other" plural format string; the second time, use the rounded number to get the plural form and select the correct pattern for that plural). If I've missed some better way of doing this, I'm open to suggestions.

Anyway, wanted to solicit feedback on this approach--thanks for looking!


rxaviers:
Hi @sieverk, it's awesome to see your WIP PR. It's on my TODO list to review and provide you feedback about your proposal. Thanks so far and sorry for the delay on my answer.

jbellenger:
@rxaviers We tried deploying @sieverk's branch at twitter and saw that some short decimal formats break the number formatter.

For example, trying to format the number 222.604 in hu ends up executing this in src/number/format.js:
```javascript
compactPattern = "000 E";  // hu cldr format
compactProperties = compactPattern.match( numberPatternRe );   // returns null
```
The cause appears to be in [numberPatternRe](https://github.com/globalizejs/globalize/blame/master/src/number/pattern-re.js#L48):
```javascript
/^(('([^']|'')*'|[^*#@0,.E])*)(\*.)?((([#,]*[0,]*0+)(\.0*[0-9]*#*)?)|([#,]*@+#*))(E\+?0+)?(('[^']+'|''|[^*#@0,.E])*)$/
```
Specifically, the last `E` in the pattern prevents matching for at least [a couple of hu number formats](https://github.com/unicode-cldr/cldr-numbers-full/blob/master/main/hu/numbers.json#L63-L68). I'll try adding some fallback behavior for when pattern matching returns null, but there may be a latent issue here with number pattern matching.

rxaviers:
`E` is a special character in the [number pattern defined by CLDR](http://www.unicode.org/reports/tr35/tr35-numbers.html#Number_Pattern_Character_Definitions).

It seems the [compact number pattern defined by CLDR](http://www.unicode.org/reports/tr35/tr35-numbers.html#Compact_Number_Formats) is slightly different in a way that the number pattern must be extracted from it.

The links I provided should contain the definition to allow that to be done correctly. If you run into any troubles please just let me know

test/unit/number/format.js
```diff
@@ -40,6 +40,14 @@ zh = new Cldr( "zh-u-nu-native" );
 
 QUnit.module( "Number Format" );
 
+function oneOrOtherPluralGenerator( plural ) {
```
rxaviers: Alternatively, we could call this `enPluralGenerator` :)

rxaviers: I see you're reusing it for `es` too, please ignore my comment.

src/number/format.js
```diff
@@ -68,6 +68,29 @@ return function( number, properties ) {
...
+			compactPattern = compactMap[ "1" + zeroes + "-count-other" ];
+			compactDigits = compactPattern.split( "0" ).length - 1;
+			divisor = zeroes.length - ( compactDigits - 1 );
+			number = number / Math.pow( 10, divisor );
+
+			// calculate rounded value to determine plural form and extract correct pattern
+			rounded = parseInt(numberFormatIntegerFractionDigits( number, minimumIntegerDigits,
+				minimumFractionDigits, maximumFractionDigits, round, roundIncrement ));
+			pluralForm = pluralGenerator ? pluralGenerator(rounded) : "other";
+			compactPattern = compactMap[ "1" + zeroes + "-count-" + pluralForm ] || compactPattern;
```
rxaviers: Please, what does `|| compactPattern` prevent? A missing `pluralForm`? Let's add a comment explaining. In theory, that should never be needed, i.e., CLDR has the necessary forms for the locale; but it's better to be safe and not throw on runtime. :+1:

src/number/format.js
```diff
@@ -68,6 +68,29 @@ return function( number, properties ) {
...
+		if ( zeroes.length >= 3 ) {
+			// use default plural form to perform initial decimal shift
+			compactPattern = compactMap[ "1" + zeroes + "-count-other" ];
+			compactDigits = compactPattern.split( "0" ).length - 1;
+			divisor = zeroes.length - ( compactDigits - 1 );
+			number = number / Math.pow( 10, divisor );
+
+			// calculate rounded value to determine plural form and extract correct pattern
+			rounded = parseInt(numberFormatIntegerFractionDigits( number, minimumIntegerDigits,
+				minimumFractionDigits, maximumFractionDigits, round, roundIncrement ));
```
rxaviers: You can drop this line and move the below code after `numberFormatSignificantDigits` and `numberFormatIntegerFractionDigits` calls, like this:

```diff
diff --git a/src/number/format.js b/src/number/format.js
index 85ed8fbe..5ca2cda2 100644
--- a/src/number/format.js
+++ b/src/number/format.js
@@ -68,26 +68,15 @@ return function( number, properties, pluralGenerator ) {
 		number *= 1000;
 	}
 
-	var compactPattern, compactDigits, compactProperties, divisor, pluralForm, rounded;
+	var compactPattern, compactDigits, compactProperties, divisor, pluralForm, zeroes;
 	if ( compactMap ) {
-		var zeroes = Array(Math.floor(number).toString().length).join( "0" );
+		zeroes = Array(Math.floor(number).toString().length).join( "0" );
 		if ( zeroes.length >= 3 ) {
 			// use default plural form to perform initial decimal shift
 			compactPattern = compactMap[ "1" + zeroes + "-count-other" ];
 			compactDigits = compactPattern.split( "0" ).length - 1;
 			divisor = zeroes.length - ( compactDigits - 1 );
 			number = number / Math.pow( 10, divisor );
-
-			// calculate rounded value to determine plural form and extract correct pattern
-			rounded = parseInt(numberFormatIntegerFractionDigits( number, minimumIntegerDigits,
-				minimumFractionDigits, maximumFractionDigits, round, roundIncrement ));
-			pluralForm = pluralGenerator ? pluralGenerator(rounded) : "other";
-			compactPattern = compactMap[ "1" + zeroes + "-count-" + pluralForm ] || compactPattern;
-			compactProperties = compactPattern.match( numberPatternRe );
-
-			// update prefix/suffix with compact prefix/suffix
-			prefix += compactProperties[ 1 ];
-			suffix = compactProperties[ 11 ] + suffix;
 		}
 	}
 
@@ -102,6 +91,16 @@ return function( number, properties, pluralGenerator ) {
 			minimumFractionDigits, maximumFractionDigits, round, roundIncrement );
 	}
 
+	if ( compactMap && zeroes.length >= 3 ) {
+		pluralForm = pluralGenerator ? pluralGenerator(+number) : "other";
+		compactPattern = compactMap[ "1" + zeroes + "-count-" + pluralForm ] || compactPattern;
+		compactProperties = compactPattern.match( numberPatternRe );
+
+		// update prefix/suffix with compact prefix/suffix
+		prefix += compactProperties[ 1 ];
+		suffix = compactProperties[ 11 ] + suffix;
+	}
+
 	// Remove the possible number minus sign
 	number = number.replace( /^-/, "" );
```

rxaviers: Such change actually fixed this:

```diff
diff --git a/test/unit/number/format.js b/test/unit/number/format.js
index ae8f9825..cd7520e0 100644
--- a/test/unit/number/format.js
+++ b/test/unit/number/format.js
@@ -120,7 +120,7 @@ QUnit.test( "decimals should format in compact mode", function( assert ) {
 	assert.equal( format( 1273, properties( "#0.#", en, { compact: "long" } ) ), "1.3 thousand" );
 	assert.equal( format( 1273000, properties( "#0.#", es, {
 		compact: "long"
-	} ), oneOrOtherPluralGenerator ), "1,3 millón" );
+	} ), oneOrOtherPluralGenerator ), "1,3 millones" );
```

test/unit/number/format.js
```diff
@@ -70,6 +78,71 @@ QUnit.test( "should format negative integer", function( assert ) {
...
+	assert.equal( format( -1273500, properties( "#0.#", en, { compact: "short" } ) ), "-1.3M" );
+	assert.equal( format( -1273500, properties( "#0.#", en, { compact: "long" } ) ), "-1.3 million" );
+	assert.equal( format( -1273500, properties( "#0.#;(#0.#)", en, { compact: "short" } ) ), "(1.3M)" );
+	assert.equal( format( -1273500, properties( "#0.#;(#0.#)", en, { compact: "long" } ) ), "(1.3 million)" );
+});
+
+QUnit.test( "percents should format in compact mode", function( assert ) {
+	assert.equal( format( 127, properties( "#0%", en, { compact: "short" } ) ), "13K%" );
+	assert.equal( format( 127, properties( "#0%", en, { compact: "long" } ) ), "13 thousand%" );
+});
```
rxaviers: Overall, let's pass `oneOrOtherPluralGenerator`. It's weird to see `1.3 million` with wrong pluralization 😄. We can add one specific test to assert the missing pluralGenerator and a missing plural form data to check both preventions you added in the code.

rxaviers:
@sieverk this is looking great, good job so far... I've left a couple of comments.

test/unit/number/format.js
rxaviers: Another tests I'd like to see are: (a) round tests, (b) significant digits tests. Example for b:

```js
 +	assert.equal( format( 1273500, properties( "@@@", en, { compact: "long" } ) ), "1.27 million" );

**Comment by rxaviers** [comment]
Something I like about the API is that significant digits can be used very conveniently to control the display digits.

```js
var fmt = .numberFormatter({
  minimumFractionDigits: 1,
  maximumFractionDigits: 2,
 compact: "short"
});
fmt( 1273500 ); // > 1.3 million
fmt( 12735000 ); // > 13 million
fmt( 127350000 ); // > 130 million

fmt = .numberFormatter({
  minimumFractionDigits: 1,
  maximumFractionDigits: 3,
  compact: "short"
});
fmt( 1273500 ); // > 1.27 million
fmt( 12735000 ); // > 12.7 million
fmt( 127350000 ); // > 127 million
```

rxaviers:
Something that seems NOT possible is to lock the scale, for example at `thousands`, but CLDR doesn't mention that possibility. By the way, the approximation in use here seems to me very in line with CLDR. I believe that on cases where user needs more control over the scale and unit, they can use formatUnit and do the math.

rxaviers:
@sieverk please could you sign the CLA? Thanks

SlexAxton:
I'm also pretty interested in getting this in. Let me know if I can help.

sieverk:
Sorry about the delayed response -- got swept up in a few other things this past month. Thanks for the feedback! I'll try to set some time aside this week to address questions / make some revisions / drop a few extra notes. Went ahead and got that CLA signed.

src/number/format.js
sieverk: Ah, nice suggestion! Lets us take full advantage of sig/decimal figure options.

src/number/format.js
sieverk: It guards against CLDR languages (Japanese, for example, but there are several) that only specify patterns for the "other" plural. In that case, checking `compactMap` for a "one" `pluralForm` would give us undefined. Hence the fallback to `compactPattern`.

test/unit/number/format.js
sieverk: Added some round and significant digits tests -- let me know if you think there could be better coverage.

Not sure what you mean about the `oneOrOtherPluralGenerator`. I'm only passing it when using `es` because it's the only language we use in the tests that has plural forms. We could pass in all of the tests, I suppose, but that will make things a bit verbose.

src/number/format.js
```diff
@@ -68,6 +68,29 @@ return function( number, properties ) {
...
+			pluralForm = pluralGenerator ? pluralGenerator(rounded) : "other";
+			compactPattern = compactMap[ "1" + zeroes + "-count-" + pluralForm ] || compactPattern;
+			compactProperties = compactPattern.match( numberPatternRe );
+
+			// update prefix/suffix with compact prefix/suffix
+			prefix += compactProperties[ 1 ];
+			suffix = compactProperties[ 11 ] + suffix;
+		}
+	}
```
sieverk: Could you elaborate on this? I'm afraid I don't quite understand what "scale" means in this context.

src/number/format.js
rxaviers: Yeap, but in theory that should never happen, for example the Japanese example, `pluralGenerator` should never return `one`. Also, CLDR has all the necessary forms for a locale. For example, for cases like Arabic where there are various plural forms, compactMap should contain all of them.

 . What I mean is,

rxaviers: EDIT: removed the garbage: "What I mean is, "

src/number/format.js
rxaviers: Sure, basically my concern is what happens when you try to format a number bigger (in orders of magnitude) than `100000000000000-count-other": "000 trillion`?

test/unit/number/format.js
rxaviers: Sure, what I mean is this:

```js
enPluralGenerator = function(n) {
  var s = String(n).split('.'), v0 = !s[1];
  return (n == 1 && v0) ? 'one' : 'other';
}

esPluralGenerator = function(n) {
  return (n == 1) ? 'one' : 'other';
}
```

rxaviers: By the way, my original message was confusing wrt that.

src/number/compact-pattern-re.js
```diff
@@ -0,0 +1,50 @@
...
+ *  7: integer_pattern
+ *  8: fraction_pattern
+ *  9: significant_pattern
+ * 10: scientific_notation
+ * 11: suffix
+ * 12: -
+ */
+return ( /^(('([^']|'')*'|[^*#@0,.E])*)(\*.)?((([#,]*[0,]*0+)(\.0*[0-9]*#*)?)|([#,]*@+#*))(E\+?0+)?(.*?)$/ );
+
+});
```
rxaviers: The diff between this and the src/number/pattern-re.js is:

```diff
--- compact-pattern-re.js	2017-06-22 11:34:01.000000000 -0300
+++ pattern-re.js	2017-03-13 11:49:00.000000000 -0300
@@ -26,7 +26,7 @@
  *
  * suffix =                   non_number_stuff
  *
- * non_number_stuff =         regexp(.*?)
+ * non_number_stuff =         regexp(('[^']+'|''|[^*#@0,.E])*)
  *
  *
  * Regexp groups:
@@ -45,6 +45,6 @@
  * 11: suffix
  * 12: -
  */
-return ( /^(('([^']|'')*'|[^*#@0,.E])*)(\*.)?((([#,]*[0,]*0+)(\.0*[0-9]*#*)?)|([#,]*@+#*))(E\+?0+)?(.*?)$/ );
+return ( /^(('([^']|'')*'|[^*#@0,.E])*)(\*.)?((([#,]*[0,]*0+)(\.0*[0-9]*#*)?)|([#,]*@+#*))(E\+?0+)?(('[^']+'|''|[^*#@0,.E])*)$/ );
 
 });
```

Can we update `src/number/pattern-re.js` instead and reuse that?

rxaviers: The definition seems a valid number pattern re.

src/number/format.js
```diff
@@ -16,11 +19,11 @@ define([
  * Return the formatted number.
  * ref: http://www.unicode.org/reports/tr35/tr35-numbers.html
  */
-return function( number, properties ) {
+return function( number, properties, pluralGenerator ) {
 	var infinitySymbol, maximumFractionDigits, maximumSignificantDigits, minimumFractionDigits,
 	minimumIntegerDigits, minimumSignificantDigits, nanSymbol, nuDigitsMap, padding, prefix,
 	primaryGroupingSize, pattern, ret, round, roundIncrement, secondaryGroupingSize, suffix,
-	symbolMap;
+	symbolMap, compactMap;
```
rxaviers: Nitpick: please sort the variables, therefore compactMap would be the first one. ☺️

rxaviers:
- [ ] Note to self: Review additional tests.

src/number/format.js
sieverk: Ahh, I see what you mean. We'll have to detect when no pattern is available for the number of digits and just return the raw number, I guess. Ideally, we'd be able to use scientific notation as a fallback when there is support for that.

test/unit/number/format.js
sieverk: Sure, I can make that change. It shouldn't affect English, though. The plural/single case is identical.

src/number/format.js
sieverk: D'oh, that's what I get for renaming too quickly.

alex-stripe:
Is there anything I can do to get this over the finish line? I'm really looking forward to using it.

rxaviers:
> Is there anything I can do to get this over the finish line? I'm really looking forward to using it.

Hi @alex-stripe, I think this is close to complete. I have to review the tests and I see [one comment above](https://github.com/globalizejs/globalize/pull/715#discussion_r123530845) not yet answered. @sieverk please any thoughts?

@alex-stripe do you use globalize-webpack-plugin or the precompilation on your side? If so, a great help would be getting those projects in good shape for this change. I believe very few will change there. Therefore, it would be mostly about adding tests to assert this works fine.

SlexAxton:
(Sorry I keep switching accounts)

I don't use either, but I'll see if I can add some tests there in the next day or so just to keep things smooth.

rxaviers:
Awesome! That will certainly make things smooth and quicker to release. Thanks

sieverk:
Yeah, I think as far as code changes, there's that issue and handling long numbers (that exceed defined patterns). Both should be pretty easy. Will see if I can set aside an hour or so today.

rxaviers:
Awesome! I see updates and I believe it's entirely on my side now :). Will review in a couple of days. Thanks!

src/number/format.js
```diff
@@ -68,6 +68,32 @@ return function( number, properties ) {
...
+	originalNumber;
+
+	// Compact mode: initial number digit processing
+	if ( compactMap ) {
+		originalNumber = number;
+		zeroes = Array( Math.floor( number ).toString().length ).join( "0" );
+		if ( zeroes.length >= 3 ) {
+
+			// use default plural form to perform initial decimal shift
+			compactPattern = compactMap[ "1" + zeroes + "-count-other" ];
```
rxaviers: At this point (the format execution itself), `compactMap` data structure should be the most convenient possible for use here (prepared by format-properties stage), e.g.:

```
compactMap: {
  3: {
    one: "0 thousand",
    other: "0 thousand"
  },
  4: {one, other},
  ...
  15: {one, other}
}
```

That would eliminate the need of creating the zeros array and would simplify the key above.

src/number/format.js
```diff
@@ -68,6 +68,32 @@ return function( number, properties ) {
...
+			if ( compactPattern ) {
+				compactDigits = compactPattern.split( "0" ).length - 1;
+				divisor = zeroes.length - ( compactDigits - 1 );
+				number = number / Math.pow( 10, divisor );
+
+				// Some languages specify no pattern for certain digit lengths, represented as "0".
+				// If no pattern, original number should remain uncompacted.
+				if ( compactPattern === "0" ) {
+					number = originalNumber;
+				}
```
rxaviers: Could this `if` block be moved up like:

```js
if ( compactPattern === "0" ) {
    number = originalNumber;
    // Could you do this and avoid entering the other if block below
    // (the one after numbers are processed)
    compactPattern = null;
} else if ( compactPattern ) {
    ...
}
```

src/number/format.js
rxaviers: Like commented below, it would be ideal to use the biggest available instead of ignoring compact mode for big numbers.

src/number/format.js
```diff
@@ -79,6 +105,22 @@ return function( number, properties ) {
...
 	}
 
+	// Compact mode: apply formatting
+	if ( compactMap && compactPattern ) {
+		pluralForm = pluralGenerator ? pluralGenerator( number ) : "other";
+		compactPattern = compactMap[ "1" + zeroes + "-count-" + pluralForm ] || compactPattern;
+
+		// Some languages specify no pattern for certain digit lengths, represented as "0".
+		// Only apply compact pattern if one is specified.
+		if ( compactPattern !== "0" ) {
```
rxaviers: Above, could you simply set `compactPattern = null` in this case and avoid this if completely?

test/unit/number/format.js
rxaviers: True, thanks

rxaviers:
I've added a couple of comments requesting changes, but I have simultaneously implemented them here: https://github.com/rxaviers/globalize/tree/compact-mode

It needs another review round...

test/unit/number/format.js
```diff
@@ -70,6 +82,123 @@ QUnit.test( "should format negative integer", function( assert ) {
...
+	// \u00A0 is a unicode non-breaking space
+	assert.equal( format( 1273, properties( "#0", hu, { compact: "short" } ) ), "1\u00A0E" );
+	assert.equal( format( 1273, properties( "#0", hu, { compact: "long" } ) ), "1 ezer" );
+	assert.equal( format(9000000, properties( "#0", hu, { compact: "short" } ) ), "9\u00A0M" );
+	assert.equal( format(9000000, properties( "#0", hu, { compact: "long" } ) ), "9 millió" );
+});
+
+QUnit.test( "unsupported numbers should apply no compacting in compact mode", function( assert ) {
+	assert.equal( format( 0.01, properties( "#0.##", en, { compact: "short" } ) ), "0.01" )
+	assert.equal( format( 1234000000000000, properties( "#0", en, { compact: "short" } ) ), "1234000000000000" )
```
sieverk: Sure, I think that's probably more in line with what the user would want.

src/number/format.js
sieverk: Ah, yes, shifting the work into compactMap makes a lot of sense.

src/number/format.js
sieverk: Yep. As you did in your branch, don't even need to apply the division / restore the number if arranged that way.

sieverk:
Looked over your changes -- they look pretty good to me, other than some minor questions. Did you want a review over there?

rxaviers:
Thanks @sieverk and sure your review is welcome. Let me create a PR.

rxaviers:
Here we go... #759

alex-stripe:
Seems like all the failures are commit message related.

Can we just rebase it to something valid and merge?

rxaviers:
Continued in #759

rxaviers:
@alex-stripe thanks for the push... It would be great if you could help by updating the examples and add a compact case if you have time. Thanks

rxaviers:
By the way, this is merged via #759...

SlexAxton:
Oh awesome. I'll whip up some examples asap.

PR #719, opened 2017-05-11
Runtime: Fix unitFormatter with numberFormatter

### LINKED ISSUE #704: Globalize-runtime cannot have two unit formatters with same options but different number formatters
Globalize version: 1.2.3
Globalize compiler version: v0.2.4

Example code (might not compile, but hopefully illustrates the point):

```ts
// compilation time - note that the only difference in formatters is the number formatter options
const unitFormatOptions1 = { numberFormatter: globalizeInstance.numberFormatter({minimumIntegerDigits: 1}) };
const unitFormatOptions2 = { numberFormatter: globalizeInstance.numberFormatter({minimumIntegerDigits: 2}) };
compileFormatters([
  globalizeInstance.unitFormatter('hour', unitFormatOptions1),
  globalizeInstance.unitFormatter('hour', unitFormatOptions2)
]);


// ... now to the runtime
const x = globalizeInstance.formatUnit(3, 'hour', unitFormatOptions1); 
const y = globalizeInstance.formatUnit(3, 'hour', unitFormatOptions2);
```
Expected: `x == "3"` and `y== "03"`
Actual: `x == "3"` and `y == "3"`

The problem is that the unit formatter takes a number formatter (a function) as one of its options, but the runtime keys are generated by `JSON.stringifying` the options ([see here](https://github.com/globalizejs/globalize/blob/c6cd869fd5a6518eab0595503be85b6b29a7b43f/src/common/runtime-key.js#L8)), and JSON.stringify stringifies functions as empty string.

So, the runtime key is resolved to the same thing in both formatters, even though they have different number formatters.

nkovacs:
JSON.stringify omits functions, so the generated runtimeKey
did not depend on the value of the numberFormatter option,
causing different unitFormatters to have an identical runtimeKey.

Fixes #704

### DIFF
--- a/src/common/runtime-stringify.js
+++ b/src/common/runtime-stringify.js
@@ -1,0 +1,12 @@
+define([], function( ) {
+
+return function( args ) {
+	return JSON.stringify( args, function( key, value ) {
+		if ( typeof value === "function" ) {
+			return value.runtimeKey; // if undefined, the value will be filtered out.
+		}
+		return value;
+	} );
+};
+
+});
--- a/src/common/runtime-bind.js
+++ b/src/common/runtime-bind.js
@@ -1,11 +1,12 @@
 define([
 	"./runtime-key",
+	"./runtime-stringify",
 	"../util/function-name"
-], function( runtimeKey, functionName ) {
+], function( runtimeKey, runtimeStringify, functionName ) {
 
 return function( args, cldr, fn, runtimeArgs ) {
 
-	var argsStr = JSON.stringify( args ),
+	var argsStr = runtimeStringify( args ),
 		fnName = functionName( fn ),
 		locale = cldr.locale;
 
--- a/src/common/runtime-key.js
+++ b/src/common/runtime-key.js
@@ -1,10 +1,11 @@
 define([
+	"./runtime-stringify",
 	"../util/string/hash"
-], function( stringHash ) {
+], function( runtimeStringify, stringHash ) {
 
 return function( fnName, locale, args, argsStr ) {
 	var hash;
-	argsStr = argsStr || JSON.stringify( args );
+	argsStr = argsStr || runtimeStringify( args );
 	hash = stringHash( fnName + locale + argsStr );
 	return hash > 0 ? "a" + hash : "b" + Math.abs( hash );
 };
--- a/test/compiler/cases/unit.js
+++ b/test/compiler/cases/unit.js
@@ -28,10 +28,9 @@ module.exports = {
 			{ formatter: Globalize.unitFormatter( "hour", {
 				numberFormatter: Globalize.numberFormatter( { minimumIntegerDigits: 1 } )
 			} ), args: [ 3 ] },
-			// TODO: Fails because of https://github.com/globalizejs/globalize/issues/704
-			/* { formatter: Globalize.unitFormatter( "hour", {
+			{ formatter: Globalize.unitFormatter( "hour", {
 				numberFormatter: Globalize.numberFormatter( { minimumIntegerDigits: 2 } )
-			} ), args: [ 3 ] } */
+			} ), args: [ 3 ] },
 
 			{ formatter: en.unitFormatter( "day" ), args: [ 1 ] },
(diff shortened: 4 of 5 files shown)


src/core-runtime.js
```diff
@@ -30,6 +30,14 @@ Globalize.locale = function( locale ) {
 	return this._locale;
 };
 
+Globalize._loadRuntimeFn = function( key ) {
+	var fn = Globalize[ key ];
+	if ( typeof fn === "function" ) {
+		fn.runtimeKey = key;
+	}
+	return fn;
+};
```
rxaviers: Oh! For your proposed solution, we need this function that is similar to [common/runtime-bind.js](https://github.com/globalizejs/globalize/blob/master/src/common/runtime-bind.js). Can you move this into its own file in `common/` and update usage to look like [this](https://github.com/globalizejs/globalize/blob/master/src/date.js#L136) please (i.e., use a function directly instead of Globalize.<static method>)?

PS: It will be challenging to find a name to distinguish this from runtime-bind. Feel free to rename the latter to accommodate a good naming for both.

rxaviers: Thinking more about it...

Instead of stamping the functions in here, we could do that in the [compiler](https://github.com/globalizejs/globalize-compiler) itself (during build time).

rxaviers:
Thanks @nkovacs!! Great catch!

I left a comment above. What do you think? Just let me know if you have any questions about it...

src/core-runtime.js
nkovacs: I wanted to avoid having to change the compiler, since it's a separate package.

rxaviers: The goal is to make the runtime code (this code) leaner, which is optimal. No worries about updating the compiler. I can help feel you if you have any questions.

nkovacs: Yeah, but how can you force a project using globalize and globalize-compiler to update the compiler when globalize is updated? If only globalize is updated, it will generate a different runtime key during compilation, and the runtime bundle won't find the correct function.

rxaviers: We cannot force, indeed, but there are ways to bring this information to the projects. One is documentation and release notes. Another is to use semver and make the existing globalize-compiler to require a peer globalize that goes up until 1.2.3; Only new globalize-compiler will support a peer globalize 1.3.0.

nkovacs: You're right, fortunately, globalize-compiler currently has a peer dependency on globalize <1.3.0, so that'll get people to upgrade the compiler as well.

rxaviers: :+1:

src/common/runtime-stringify.js
```diff
@@ -0,0 +1,12 @@
+define([], function( ) {
+
+return function( args ) {
+	return JSON.stringify( args, function( key, value ) {
+		if ( typeof value === "function" ) {
+			return value.runtimeKey; // if undefined, the value will be filtered out.
```
nkovacs: Not sure if we should do something about functions that don't have a runtimeKey.

nkovacs:
Pull request for compiler is up: https://github.com/globalizejs/globalize-compiler/pull/31
and I've removed the hack in core-runtime.js.

nkovacs:
I've rebased on master and uncommented the compiler test. Travis fails because of the missing compiler change.

rxaviers:
Ok thanks

rxaviers:
@nkovacs, other formatters (e.g., date, relative time) also include a number formatter in their runtime properties. We need to understand why this happens with unit formatter, but not with the other ones...

nkovacs:
No other formatter accepts a formatter as an option, which should affect the runtimeKey.

rxaviers:
True! Thanks

rxaviers:
Thanks @nkovacs!

> Pull request for compiler is up: globalizejs/globalize-compiler#31
and I've removed the hack in core-runtime.js.

Instead of depending on globalize-compiler to include the runtimeKey info, I've update the runtime function to [do it](https://github.com/globalizejs/globalize/commit/a9754b630f2f16494abbe5def057e9875225f2e1#diff-95c9e38f0b1ac4fc38bb6b17833f4864R37).

PR #721, opened 2017-05-14
Test: add tests for compiler

nkovacs:
globalize-compiler complains about missing peer dependencies. One is globalize, the other two are cldr-data and iana-tz-data.

Globalize is obviously available, cldr-data is also available from bower. So it still works.

iana-tz-data was recently added to globalize-compiler, but it's not in globalize.js's master, only in the 1.3.0-alpha.3 tag. I think you might have forgotten to push to master.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -79,7 +79,7 @@ module.exports = function( grunt ) {
 			},
 			test: {
 				src: [ "test/*.js", "test/functional/**/*.js", "test/unit/**/*.js",
-					"!test/config.js" ],
+					"test/compiler/**/*.js", "!test/config.js", "!test/compiler/_compiled/**" ],
 				options: {
 					jshintrc: "test/.jshintrc"
 				}
@@ -94,8 +94,17 @@ module.exports = function( grunt ) {
 		jscs: {
 			source: [ "src/**/*.js", "!src/build/**" ],
 			grunt: "Gruntfile.js",
-			test: [ "test/*.js", "test/functional/**/*.js", "test/unit/**/*.js" ],
+			test: [ "test/*.js", "test/functional/**/*.js", "test/unit/**/*.js",
+				"test/compiler/**/*.js", "!test/compiler/_compiled/**" ],
 			dist: [ "dist/globalize*.js", "dist/globalize/*.js" ]
+		},
+		mochaTest: {
+			test: {
+				options: {
+					reporter: "spec"
+				},
+				src: [ "test/compiler/*.js" ]
+			}
 		},
 		qunit: {
 			functional: {
@@ -655,6 +664,7 @@ module.exports = function( grunt ) {
 		// TODO fix issues, enable
 		// "jscs:dist",
 		"test:functional",
+		"mochaTest",
 		"uglify",
 		"compare_size",
 		"commitplease"
--- a/test/.jshintrc
+++ b/test/.jshintrc
@@ -16,6 +16,10 @@
 	"globals": {
 		"define": false,
 		"require": false,
+		"module": false,
+		"__dirname": false,
+		"describe": false,
+		"it": false,
 		"QUnit": false
 	}
 }
--- a/test/compiler/test.js
+++ b/test/compiler/test.js
@@ -1,0 +1,74 @@
+var globalizeCompiler = require( "globalize-compiler" );
+var fs = require( "fs" );
+var path = require( "path" );
+var childProcess = require( "child_process" );
+var assert = require( "assert" );
+var glob = require( "glob" );
+
+describe( "compiled", function() {
+	if ( !fs.existsSync( path.join( __dirname, "_compiled" ) ) ) {
+		fs.mkdirSync( path.join( __dirname, "_compiled" ) );
+	}
+	var files = glob.sync( "cases/*.js", {
+		cwd: __dirname
+	} );
+	files.forEach( function(file) {
+		var name = path.basename( file, ".js" );
+		describe( name, function() {
+			it( "should return identical data", function( done ) {
+				var test = require( "./" + file );
+				var expectedOutput = "";
+				var formatters = [];
+				var testCases = test.cases( test.dependencies() );
+				testCases.forEach( function( testCase ) {
+					formatters.push( testCase.formatter );
+					expectedOutput += testCase.formatter.apply( null, testCase.args ) + "\n";
+				});
+				var out = globalizeCompiler.compile( formatters, {
+					template: function( data ) {
+						var deps = "var Globalize = " +
+							data.dependencies.map( function( dependency ) {
+								return "require( \"../../../dist/" + dependency + "\" )";
+							} ).join( ";\n" );
+						return [
+							deps,
+							"",
+							data.code,
+							"",
+							"module.exports = Globalize;"
+						].join( "\n" );
+					}
+				} );
+				fs.writeFileSync( path.join( __dirname, "_compiled/" + name + ".compiled.js" ), out );
+				fs.writeFileSync( path.join( __dirname, "_compiled/" + name + ".test.js" ), [
+					"var test = require( \"../" + file + "\" );",
+					"var Globalize = require( \"../_compiled/" + name + ".compiled.js\" );",
+					"testCases = test.cases( Globalize );",
+					"testCases.forEach( function( testCase ) {",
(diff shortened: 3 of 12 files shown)


Gruntfile.js
```diff
@@ -79,7 +79,7 @@ module.exports = function( grunt ) {
 			},
 			test: {
 				src: [ "test/*.js", "test/functional/**/*.js", "test/unit/**/*.js",
-					"!test/config.js" ],
+					"test/compiler/**/*.js", "!test/config.js", "!test/compiler/_compiled/**" ],
```
rxaviers: Let's generate the output somewhere else outside of `test/`. Can we? Therefore, we don't need the negatives below for linters.

test/.jshintrc
```diff
@@ -16,6 +16,10 @@
 	"globals": {
 		"define": false,
 		"require": false,
+		"module": false,
+		"__dirname": false,
+		"describe": false,
+		"it": false,
```
rxaviers: Let's move your just created `test/compiler` into `test-compiler`. That way we isolate the existing tests that use QUnit and your just created ones that use mocha.

rxaviers:
Hi @nkovacs, this is a great addition, thank you!

Having said that, there is a detail I would prefer done differently: the tests should be explicit, not dynamically generated. Currently, the `it`s are generated dynamically by reading `./cases` data. While I understand such approach leads into smaller test file sizes and to allow generating the compiled files, but for tests it's more important they are easily readable/understood than those.

Please, could you update that? Globalize compiler tests could be used as baseline (note its functional tests) https://github.com/globalizejs/globalize-compiler

Thanks

Gruntfile.js
nkovacs: I think it has to be inside the working directory (so it can't be `os.tmpdir()`), because it needs to find `node_modules`.

rxaviers: I mean a sibling to `test/` and `node_modules/`.

nkovacs:
The problem is that when the runtime bundle is compiled, it needs to be executed with the exact same formatters. That's why I need the separate case files.

Some pseudocode:

```
describe("The compiled `basic.js`", function() {
	it("should include formatDate", function() {
		var formatter = Globalize.dateFormatter({date: "medium"});
		var expected = formatter(new Date(2017, 3, 15));

		compiled = compile(formatter); // This is just the runtime bundle
		// Write compiled to a file.

		// Now I need to generate a file that requires the compiled bundle and contains
		// console.log(Globalize.dateFormatter({date: "medium"})(new Date(2017, 3, 15)));
		// How do I do that without duplicating everything?

		// Then run this file and get its output, put it into actual.

		assert.equal( actual, expected );
	});
});
```

The only thing I can think of here is to put the formatter part into a function, and use function.toString() to generate the testfile.
E.g.:

```
describe("The compiled `basic.js`", function() {
	it("should include formatDate", function() {
		var formatter;
		var testFunction = function() {
			formatter = Globalize.dateFormatter({date: "medium"});
			return formatter(new Date(2017, 3, 15));
		};
		var expected = testFunction();

		compiled = compile(formatter); // This is just the runtime bundle
		// Write compiled to a file.

		var code = "require('_tmp/compiled.js'); var testFunction = " +
			testFunction().toString() +
			"; console.log(testFunction)";

		// Write code to a file.
		// Then run this file and get its output, put it into actual.

		assert.equal( actual, expected );
	});
});
```

I prefer the current code because you have the different test cases clearly listed in the testcase files without too much overhead.

rxaviers:
Or the compiled code is generated before the tests are executed, which is the approach taken in globalize-compiler tests: https://github.com/globalizejs/globalize-compiler

Gruntfile.js
nkovacs: But why not keep everything related to these tests in the same subdirectory? In globalize-compiler's tests the fixtures are inside the tests/functional directory too.

rxaviers: > But why not keep everything related to these tests in the same subdirectory

Because, they are generated, which means we aren't going to lint them nor to commit them. Keeping these files in a separate directory makes that easier.

> In globalize-compiler's tests the fixtures are inside the tests/functional directory too.

... which I plan to fix there as well.

nkovacs: It also pollutes your working directory. Instead of having a single `test` directory, you'll now have `test`, `test-compiler` and `test-compiler-generated`.

rxaviers: For the generated compiled thing, we can use the existing `tmp/`. For the tests, let's use the two separate directories for now. Possibly we could consolidate the test framework to be mocha and merge both directories back into one. For now, I want to avoid having mixed content, see the unnecessary jshintrc additions for qunit perspective.

nkovacs:
In globalize-compiler the fixtures are generated before the tests run. This means that technically you are not actually testing the compiler, you're just testing the runtime code, and assuming that the bundle is always generated correctly. You don't get proper code coverage, you can't use watch mode, and you can't run mocha directly either.

In globalize-compiler's functional tests, the only reason [this test](https://github.com/globalizejs/globalize-compiler/blob/master/test/functional/index.js#L7) works, for example, is that the fixture contains a completely unrelated [Globalize.formatDate](https://github.com/globalizejs/globalize-compiler/blob/master/test/unit/fixtures/basic.js#L5) call. This dependency is not obvious and easy to break, especially since the functional tests use the fixtures of the unit tests, it's not obvious that those fixtures are needed by the functional tests.

The tests in globalize-compiler also validate the exact output. The tests here do not. They only check that the testcase has the exact same output when running with the runtime bundle as when running with the full fat globalize.js + cldr data, because that's what you care about when testing the compiler. You already have tests that verify that the individual formatters do what they're supposed to do, there's no need to check that here too, and it makes maintenance simpler.

While I would prefer to be able to have explicit `its` instead of dynamically generating them, there's [nothing wrong with doing that](https://mochajs.org/#dynamically-generating-tests). [Pugjs does something similar](https://github.com/pugjs/pug/blob/master/packages/pug/test/run.test.js), and I used that as inspiration.
While you lose some clarity in `test.js`, you gain that in the `cases/*.js` files, since you have all the formatters that are being tested in a reasonably readable form, with no duplication.

rxaviers:
You have valid points! :+1: cc @jzaefferer

test/compiler/test.js
```diff
@@ -0,0 +1,74 @@
...
+	if ( !fs.existsSync( path.join( __dirname, "_compiled" ) ) ) {
+		fs.mkdirSync( path.join( __dirname, "_compiled" ) );
+	}
+	var files = glob.sync( "cases/*.js", {
+		cwd: __dirname
+	} );
+	files.forEach( function(file) {
+		var name = path.basename( file, ".js" );
+		describe( name, function() {
+			it( "should return identical data", function( done ) {
```
rxaviers: should return identical output compared to normal globalize? Can we update the description

Gruntfile.js
nkovacs: If it's only the jshintc additions that are bothering you, another option is to use `extends` (assuming that works properly, I've never tried it, and it's not documented very well).
Or to switch to eslint, which also supports extending config files in subdirectories.
But I'll separate the directories for now.

rxaviers: > Or to switch to eslint

We definitely should have moved to eslint by now... :smile: 

> But I'll separate the directories for now.

Rethinking about it, feel free to ignore it. If you have done it already, ok; otherwise lets leave it as is.

rxaviers: The existing test stack (grunt, AMD, phantomjs, QUnit) is something I wanted to get updated to using npm scripts, mocha only. So, your tests are closer to that.

nkovacs: Will putting the generated files into `tmp/` affect the build/publish process? That's where grunt-contrib-uglify puts the minified files, but it appears they are not used after that.

rxaviers: When the build process finishes, `tmp/.build/` files are saved into `dist/` and `tmp/` is left/abandoned.

rxaviers:
Merged! Thank you very much @nkovacs. This is a great already and improvements could be done later.

test/compiler/test.js
nkovacs: Sorry, I didn't have time to work on this. Do you still want me to do this? Or you can just commit it directly to master.

rxaviers: No worries, this can be done later.

PR #730, opened 2017-05-23
Date: Loose Matching

### LINKED ISSUE #292: Parse: Loose Matching

rxaviers:
We have loose matching #292 on number parsing. This change applies it to date parsing as well. See tests for examples.

Fixes #292

### DIFF
--- a/src/build/intro-date-runtime.js
+++ b/src/build/intro-date-runtime.js
@@ -42,6 +42,7 @@
 "use strict";
 
 var createErrorUnsupportedFeature = Globalize._createErrorUnsupportedFeature,
+	looseMatching = Globalize._looseMatching,
 	regexpEscape = Globalize._regexpEscape,
 	removeLiteralQuotes = Globalize._removeLiteralQuotes,
 	runtimeKey = Globalize._runtimeKey,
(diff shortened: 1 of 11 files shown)


rxaviers:
We have* (edit)

Issue #720, opened 2017-05-12
version check for compiled bundles


nkovacs:
#719 and most likely any fix for #705 would make existing runtime bundles incompatible with 1.3.0's runtime code. If you update your npm packages, but forget to rebuild the bundle, your code will break.

I don't know how likely this is, but to be safe, my idea is to add a version number to the compiled bundle, and warn the user if the version number is incompatible.

This version number would be an integer, which is incremented whenever there's a breaking change. If the version number is missing, it'll default to 0, and so also cause a warning.

What do you think?

rxaviers:
I think it's a great idea. Ideally the used version number can be deduced from package.json's version.

nkovacs:
So, every new minor version would be incompatible?

rxaviers:
I would explore something in this line:
- Globalize could expose its own version via `Globalize.version` attribute.
- The compiled bundle could have an init check that asserts if Globalize.version is within a certain range. We need to use a very lightweight approach here.
- Globalize-compiler should be able to generate the init checker based on its peer dependency definition over globalize.

We need to explore ideas.

nkovacs:
My original idea was to just have a simple integer, and the runtime check would be a `===`. This is simpler than a semver check.
This integer would be defined in globalize.js, e.g. `Globalize.runtimeVersion`, the compiler would read it and generate something like this (hard-coding the integer into the bundle):

```
Globalize._checkRuntimeVersion(12);
```

`_checkRuntimeVersion` could be as simple as:

```
function(bundleVersion) {
    if (bundleVersion !== Globalize.runtimeVersion) {
        console.warn('Version mismatch etc.');
    }
}
```

Globalize.runtimeVersion could be generated automatically when building globalize.js, e.g. major*100+minor, or it could be handled manually, when you know that you broke compatibility.

Unfortunately, existing bundles wouldn't be checked with this solution.

rxaviers:
Closing this issue, because https://github.com/globalizejs/globalize/pull/719#issuecomment-404961144 was solved using an approach that doesn't require this.

rxaviers:
Thanks!

PR #754, opened 2017-07-04
Runtime: add runtime version, set it to 1

nkovacs:
Refs #720 

The corresponding compiler PR is https://github.com/globalizejs/globalize-compiler/pull/34

I did it this way so that if you're directly using compiler.compile, it'll work without changes.
The compiler will also work without the change in globalize.js.

### DIFF
--- a/src/common/runtime-bind.js
+++ b/src/common/runtime-bind.js
@@ -1,7 +1,8 @@
 define([
 	"./runtime-key",
-	"../util/function-name"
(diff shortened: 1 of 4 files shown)


rxaviers:
I'm closing this PR, because https://github.com/globalizejs/globalize/pull/719#issuecomment-404961144 was solved using an approach that doesn't require this. Thanks a lot and sorry for the delay.

PR #763, opened 2017-08-03
Date: Add fractional seconds support in formatters

### LINKED ISSUE #753: 1.3-rc0 fractional seconds nor milliseconds working for date skeletons
1.3-rc is looking good, but this is blocking us right now.

Repro!
```
const ianaTzData = require('iana-tz-data');
const Globalize = require("globalize");
const cldrData = require('./vendor/cldr-data.json');
Globalize.loadTimeZone(ianaTzData);
Globalize.load(cldrData);
Globalize.locale("en");

const date = new Date();
const d1 = Globalize.dateFormatter({ skeleton: "yyyyMMddjjmmssSSS" })
console.log(d1(date))
const d2 = Globalize.dateFormatter({ skeleton: "yyyyMMddjjmmssAAA" })
console.log(d2(date))
```
Expected to see fractional seconds and milliseconds according to this: http://www.unicode.org/reports/tr35/tr35-dates.html#dfst-second

Actual Output:
```
D:\gitlab\localize> node .\test.js
06/30/2017, 02:00:00037 PM
06/30/2017
```

mattyork:
Follows this algorithm:
http://www.unicode.org/reports/tr35/tr35-dates.html#Matching_Skeletons

Fixes #753

### DIFF
--- a/src/date/expand-pattern/augment-format.js
+++ b/src/date/expand-pattern/augment-format.js
@@ -4,7 +4,7 @@ define([
 	"../../util/string/repeat"
 ], function( dateExpandPatternNormalizePatternType, datePatternRe, stringRepeat ) {
 
-function expandBestMatchFormat(skeletonWithoutFractionalSeconds, bestMatchFormat) {
+function expandBestMatchFormat( skeletonWithoutFractionalSeconds, bestMatchFormat ) {
 	var i, j, bestMatchFormatParts, matchedType, matchedLength, requestedType,
 		requestedLength, requestedSkeletonParts,
 
@@ -33,22 +33,22 @@ function expandBestMatchFormat(skeletonWithoutFractionalSeconds, bestMatchFormat
 
 // See: http://www.unicode.org/reports/tr35/tr35-dates.html#Matching_Skeletons
 return function( requestedSkeleton, bestMatchFormat, decimalSeparator ) {
-	var countOfFractionalSeconds, fractionalSecondMatch,lastSecondIdx,
-		skeletonWithoutFractionalSeconds,
+	var countOfFractionalSeconds, fractionalSecondMatch, lastSecondIdx,
+		skeletonWithoutFractionalSeconds;
 
 	fractionalSecondMatch = requestedSkeleton.match( /S/g );
 	countOfFractionalSeconds = fractionalSecondMatch ? fractionalSecondMatch.length : 0;
 	skeletonWithoutFractionalSeconds = requestedSkeleton.replace( /S/g, "" );
 
-	bestMatchFormat = expandBestMatchFormat(skeletonWithoutFractionalSeconds, bestMatchFormat);
+	bestMatchFormat = expandBestMatchFormat( skeletonWithoutFractionalSeconds, bestMatchFormat );
 
-	lastSecondIdx = bestMatchFormat.lastIndexOf("s");
-	if (lastSecondIdx !== -1 && countOfFractionalSeconds !== 0 ) {
+	lastSecondIdx = bestMatchFormat.lastIndexOf( "s" );
+	if ( lastSecondIdx !== -1 && countOfFractionalSeconds !== 0 ) {
 		bestMatchFormat =
-			bestMatchFormat.slice(0, lastSecondIdx + 1) +
+			bestMatchFormat.slice( 0, lastSecondIdx + 1 ) +
 			decimalSeparator +
 			stringRepeat( "S", countOfFractionalSeconds ) +
-			bestMatchFormat.slice(lastSecondIdx + 1);
+			bestMatchFormat.slice( lastSecondIdx + 1 );
 	}
 	return bestMatchFormat;
 };
--- a/src/date/expand-pattern/augment-format.js
+++ b/src/date/expand-pattern/augment-format.js
@@ -4,30 +4,42 @@ define([
 	"../../util/string/repeat"
 ], function( dateExpandPatternNormalizePatternType, datePatternRe, stringRepeat ) {
 
-return function( requestedSkeleton, bestMatchFormat ) {
-	var i, j, matchedType, matchedLength, requestedType, requestedLength,
+// See: http://www.unicode.org/reports/tr35/tr35-dates.html#Matching_Skeletons
+return function( requestedSkeleton, bestMatchFormat, decimalSeparator ) {
+	var i, j, bestMatchFormatParts, countOfFractionalSeconds, fractionalSecondMatch, matchedType,
+		matchedLength, requestedType, requestedLength, requestedSkeletonParts,
+		skeletonWithoutFractionalSeconds,
 
 		// Using an easier to read variable.
 		normalizePatternType = dateExpandPatternNormalizePatternType;
 
-	requestedSkeleton = requestedSkeleton.match( datePatternRe );
-	bestMatchFormat = bestMatchFormat.match( datePatternRe );
+	fractionalSecondMatch = requestedSkeleton.match( /S/g );
+	countOfFractionalSeconds = fractionalSecondMatch ? fractionalSecondMatch.length : 0;
+	skeletonWithoutFractionalSeconds = requestedSkeleton.replace( /S/g, "" );
 
-	for ( i = 0; i < bestMatchFormat.length; i++ ) {
-		matchedType = bestMatchFormat[i].charAt( 0 );
-		matchedLength = bestMatchFormat[i].length;
-		for ( j = 0; j < requestedSkeleton.length; j++ ) {
-			requestedType = requestedSkeleton[j].charAt( 0 );
-			requestedLength = requestedSkeleton[j].length;
+	requestedSkeletonParts = skeletonWithoutFractionalSeconds.match( datePatternRe );
+	bestMatchFormatParts = bestMatchFormat.match( datePatternRe );
+
+	for ( i = 0; i < bestMatchFormatParts.length; i++ ) {
+		matchedType = bestMatchFormatParts[i].charAt( 0 );
+		matchedLength = bestMatchFormatParts[i].length;
+		for ( j = 0; j < requestedSkeletonParts.length; j++ ) {
+			requestedType = requestedSkeletonParts[j].charAt( 0 );
+			requestedLength = requestedSkeletonParts[j].length;
 			if ( normalizePatternType( matchedType ) === normalizePatternType( requestedType ) &&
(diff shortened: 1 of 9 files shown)


mattyork:
I signed the CLA and added the email address used for the commit to my GitHub account, but it's still not updated. Help?

test/compiler/cases/date.js
```diff
@@ -29,7 +29,7 @@ module.exports = {
 
 			{ formatter: Globalize.dateFormatter({ skeleton: "GyMMMEd" }), args: [ date ] },
 			{ formatter: Globalize.dateFormatter({ skeleton: "dhms" }), args: [ date ] },
-			{ formatter: Globalize.dateFormatter({ skeleton: "GyMMMEdhms" }), args: [ date ] },
+			{ formatter: Globalize.dateFormatter({ skeleton: "GyMMMEdhmsSSS" }), args: [ date ] },
```
rxaviers: Let's keep the original test without `S` and have the additional test with `S`.

src/date/expand-pattern/augment-format.js
```diff
@@ -4,30 +4,42 @@ define([
...
 		}
+
+		if ( matchedType === "s" && countOfFractionalSeconds !== 0 ) {
+			bestMatchFormatParts[i] +=
+				decimalSeparator + stringRepeat( "S", countOfFractionalSeconds );
+		}
 	}
 
-	return bestMatchFormat.join( "" );
+	return bestMatchFormatParts.join( "" );
```
rxaviers: Good name addition, i.e., `Parts` which they are.

src/date/expand-pattern/augment-format.js
```diff
@@ -4,30 +4,42 @@ define([
...
+	countOfFractionalSeconds = fractionalSecondMatch ? fractionalSecondMatch.length : 0;
+	skeletonWithoutFractionalSeconds = requestedSkeleton.replace( /S/g, "" );
 
-	for ( i = 0; i < bestMatchFormat.length; i++ ) {
-		matchedType = bestMatchFormat[i].charAt( 0 );
-		matchedLength = bestMatchFormat[i].length;
-		for ( j = 0; j < requestedSkeleton.length; j++ ) {
-			requestedType = requestedSkeleton[j].charAt( 0 );
-			requestedLength = requestedSkeleton[j].length;
+	requestedSkeletonParts = skeletonWithoutFractionalSeconds.match( datePatternRe );
```
rxaviers: I'm wondering how this approach scales... Maybe, not worth worrying about it now, but for example, I'm wondering if handling timeZone would take a similar approach.

rxaviers: Does it make sense to leave this function mostly as is, and handle `S` by a parent function? Basically, by replacing the existing call stack:

```diff
- getBestMatchPattern() -> augmentFormat()
+ getBestMatchPattern() -> stuffThatHandlesS() -> augmentFormat()
```

test/unit/date/expand-pattern/get-best-match-pattern.js
```diff
@@ -30,7 +30,7 @@ QUnit.test( "should get best match pattern", function( assert ) {
 	assert.equal( getBestMatchPattern( en, "cccd" ), "d EEE" );
 	assert.equal( getBestMatchPattern( en, "ccccd" ), "d EEEE" );
 
-	assert.equal( getBestMatchPattern( en, "hhmms" ), "hh:mm:ss a" );
+	assert.equal( getBestMatchPattern( en, "hhmmsS" ), "hh:mm:ss.S a" );
```
rxaviers: Can we preserve the original test and have yours as an additional one?

test/compiler/cases/date.js
mattyork: sounds good!

test/unit/date/expand-pattern/get-best-match-pattern.js
mattyork: Yep yep!

src/date/expand-pattern/augment-format.js
mattyork: Yeah, that makes sense. Is it cool if I keep both of those functions in `augment-format.js`? That way the file perfectly reflects the "adjustment" section of the [spec](http://www.unicode.org/reports/tr35/tr35-dates.html#Missing_Skeleton_Fields): 

> Once a skeleton match is found, the corresponding pattern is used, but with adjustments. Consider the following dateFormatItem:
> 
>     <dateFormatItem id="yMMMd">d MMM y</dateFormatItem>
> If this is the best match for yMMMMd, pattern is automatically expanded to produce the pattern "d MMMM y" in response to the request. Of course, if the desired behavior is that a request for yMMMMd should produce something other than "d MMMM y", a separate dateFormatItem must be present, for example:
> 
>     <dateFormatItem id="yMMMMd">d 'de' MMMM 'de' y</dateFormatItem>
> If the requested skeleton included both seconds and fractional seconds and the dateFormatItem skeleton included seconds but not fractional seconds, then the seconds field of the corresponding pattern should be adjusted by appending the locale’s decimal separator, followed by the sequence of ‘S’ characters from the requested skeleton.

src/date/expand-pattern/augment-format.js
```diff
@@ -4,30 +4,53 @@ define([
...
 			}
 		}
 	}
 
-	return bestMatchFormat.join( "" );
+	return bestMatchFormatParts.join( "" );
+}
+
+// See: http://www.unicode.org/reports/tr35/tr35-dates.html#Matching_Skeletons
+return function( requestedSkeleton, bestMatchFormat, decimalSeparator ) {
```
mattyork: @rxaviers how about this? Same file, two functions. Encapsulates all the "adjustments" to the `bestMatchFormat`

rxaviers: It's good except for our build process, which assumes all files have one export (and no outer code). In order to have something different, we need an update the [onBuildWrite](https://github.com/globalizejs/globalize/blob/master/Gruntfile.js#L153), specifically tell it this file is a "[type b](https://github.com/globalizejs/globalize/blob/master/Gruntfile.js#L282)".

mattyork: Got it. There's a problem though: if I make it a "type b", that capitalizes the first character of the variable name, which breaks the output, as the actual variable name does not have a capital first letter. Example output: 

```
var DateExpandPatternAugmentFormat = (function() {
function expandBestMatchFormat( skeletonWithoutFractionalSeconds, bestMatchFormat ) {
    ...
}

// See: http://www.unicode.org/reports/tr35/tr35-dates.html#Matching_Skeletons
return function( requestedSkeleton, bestMatchFormat, decimalSeparator ) {
    ...
    bestMatchFormat = expandBestMatchFormat( skeletonWithoutFractionalSeconds, bestMatchFormat );
    ...
};

}());
```

Why does it capitalize the first letter in `DateExpandPatternAugmentFormat`? The only other "type b" is "util/globalize-date", but that file doesn't seem to exist anymore, so I can't look at the example to know what the intention of the capitalization is. 

However, if I leave it as a "type a", everything still works. This is the output:

```
function expandBestMatchFormat( skeletonWithoutFractionalSeconds, bestMatchFormat ) {
    ...
}

var dateExpandPatternAugmentFormat = function( requestedSkeleton, bestMatchFormat, decimalSeparator ) {
    ...
    bestMatchFormat = expandBestMatchFormat( skeletonWithoutFractionalSeconds, bestMatchFormat );
    ...
};
```

This works just fine. 

So, what should I do here? 
1. Leave as a `type a`
2. Change how the build handles `type b` so it stops capitalizing. If I do this should I also remove the reference to `util/globalize-date`?
3. Something else? Move the other function to another file?

mattyork: @rxaviers ?

rxaviers: Number 2. (sorry for the delay)

Yeap, existing "type b" is bugged. Let's fix it by stop capitalizing the variable and by also removing the reference to the non-existing `util/globalize-date`. Please, take a look at this old-but-correct "type b" implementation: [here](https://github.com/globalizejs/globalize/blob/6e4de14350945b0a6674e4316ed25c58dd700403/Gruntfile.js#L126-L129) and [here](https://github.com/globalizejs/globalize/blob/6e4de14350945b0a6674e4316ed25c58dd700403/src/date/parse.js).

Thanks!

mattyork: Done!

mattyork:
How bout meow?

test/functional/date/date-formatter.js
```diff
@@ -88,7 +88,7 @@ QUnit.test( "should return a formatter", function( assert ) {
 
 	assert.equal( Globalize.dateFormatter({ skeleton: "GyMMMEd" })( date ), "Wed, Sep 15, 2010 AD" );
 	assert.equal( Globalize.dateFormatter({ skeleton: "dhms" })( date ), "15, 5:35:07 PM" );
-	assert.equal( Globalize.dateFormatter({ skeleton: "GyMMMEdhms" })( date ), "Wed, Sep 15, 2010 AD, 5:35:07 PM" );
+	assert.equal( Globalize.dateFormatter({ skeleton: "GyMMMEdhmsSSS" })( date ), "Wed, Sep 15, 2010 AD, 5:35:07.369 PM" );
```
rxaviers: Let's not replace the above test, but add this as a new one.

test/functional/date/parse-date.js
```diff
@@ -141,8 +141,8 @@ QUnit.test( "should parse skeleton", function( assert ) {
 	date.setHours( 17 );
 	date.setMinutes( 35 );
 	date.setSeconds( 7 );
-	date = startOf( date, "second" );
-	assertParseDate( assert, "Wed 5:35:07 PM", { skeleton: "Ehms" }, date );
+	date.setMilliseconds(734);
+	assertParseDate( assert, "Wed 5:35:07.734 PM", { skeleton: "EhmsSSS" }, date );
```
rxaviers: Same here...

rxaviers:
Great @mattyork, one more test update and I think this is it.

mattyork:
Aye aye, tests updated

mattyork:
Poke

mattyork:
@rxaviers ?

rxaviers:
Hi @mattyork, I didn't forget about you. Sorry about the delay. I will merge your PR. I hope the delay doesn't impact any of your work.

PR #768, opened 2017-08-17
Examples: Make basic globalize-compiler example more generic

### LINKED ISSUE #637: Make basic globalize-compiler example more generic
In `app.js`, the default locale is set with `Globalize.locale('en')`. That should happen on development only, therefore we could move that statement into `development.html`.

The default locale in production should be set by the compiler so it can generate production bundles for various locales.

jacalata:
Examples

Move  Globalize.locale('en') to development.html

Fixes #637

### DIFF
--- a/examples/globalize-compiler/app.js
+++ b/examples/globalize-compiler/app.js
@@ -1,8 +1,5 @@
 var like, number;
 
-// Set default locale as "en".
-Globalize.locale( "en" );
-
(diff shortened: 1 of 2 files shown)


jacalata:
updated my email, should fix the CLA

jacalata:
CI build failed because of my commit title I think, updated it.

rxaviers:
Thanks

PR #779, opened 2017-10-15
Add jsDelivr hits badge

LukasDrgon:
Hey, I noticed that globalize gets 3M hits per month on jsDelivr, so I thought you might like this badge.


rxaviers:
@jzaefferer @scottgonzalez any comments here? I'm afraid including one cdn badge here opens the door for other cnds to be included too.

scottgonzalez:
I'm not a big fan of CDNs or badges (there I go again being against popular things in the JS community). For JS libraries, CDNs typically provide much less value than developers expect. Badges generally just provide fluff or misleading information. In this specific case, 3M hits from jsDeliver does not tell me anything about the overlap of versions and users for those sites providing the hits. This may still result in zero cache hit benefit while adding a DNS lookup and potentially extra bytes that I don't need (because it's not customized properly). The only benefit at that point is that you're (hopefully) hitting a closer server for the file, but if you're big enough you've probably got a CDN for all of your custom JS anyway.

So, to answer your question: I wouldn't add it.

rxaviers:
@LukasDrgon thanks for your suggestion to include this CDN badge, but for the reasons above, we'll abstain from including it.

PR #780, opened 2017-10-17
LICENSE: Add MIT header

rxaviers:
@scottgonzalez @kborchers please can you confirm this update is ok? Thanks

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright jQuery Foundation and other contributors, https://jquery.org/
+Copyright JS Foundation and other contributors, https://js.foundation
 
 This software consists of voluntary contributions made by many
 individuals. For exact contribution history, see the revision history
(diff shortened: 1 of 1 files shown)


kborchers:
Not sure how we missed this before but the copyright line needs to be updated to

`Copyright JS Foundation and other contributors, https://js.foundation`

Otherwise the addition of the MIT header LGTM

rxaviers:
> Copyright JS Foundation and other contributors, https://js.foundation

Updated in https://github.com/rxaviers/globalize/tree/license-mit, though it seems like github hasn't synced my updates in this PR yet (weird, never saw that before)

kborchers:
Wow, that is very strange 😕

rxaviers:
There we go, now it's synced.

scottgonzalez:
I know this was already merged, but since I was tagged, I figured I'd still leave my feedback.

I personally avoid doing this type of labeling because we've gone through the effort of informing users that even though this package itself is released under the terms of the MIT license, it has portions of its own code, as well as dependencies, that are released under different licenses. If you want to include the heading of "The MIT License (MIT)", I would suggest adding that on L14 rather than L1.

rxaviers:
Makes sense Scott, I'm pushing an amend commit soon.

rxaviers:
Updated in d01e049

PR #795, opened 2017-12-28
Currency: currency parsing

goatandsheep:
Adding [currency parsing](https://github.com/globalizejs/globalize/issues/364) to currency.js by:

1. Parse the currency string for numbers, commas, and periods: `valueString.replace(/[^\d\.\,]/g, '')`
2. Run `numberParser`

Currency codes define the symbols and locale defines the number format. If we're just parsing the number, we can strip away the symbols and parse the number in the locale specified in the settings

### DIFF
--- a/src/currency.js
+++ b/src/currency.js
@@ -149,10 +149,11 @@ Globalize.parseCurrency =
 Globalize.prototype.parseCurrency = function( value, options ) {
 	validateParameterPresence( value, "value" );
 
-	value = value.replace(/[^\d\.\,]/g, "");
-	value = this.numberParser(value);
+	value = value.replace( /[^\d\.\,]/g, "" );
+	value = this.numberParser( value );
 
 	this.currencyParser( options )( value );
+
 	return value;
 };
 
--- a/src/currency.js
+++ b/src/currency.js
@@ -114,7 +114,7 @@ Globalize.prototype.currencyParser = function( options ) {
 	// returnFn = numberParser(value)
 
 	runtimeBind( args, returnFn, [ numberParser, properties ] );
-	return returnFn
+	return returnFn;
 };
 
 /**
@@ -149,10 +149,10 @@ Globalize.parseCurrency =
 Globalize.prototype.parseCurrency = function( value, options ) {
 	validateParameterPresence( value, "value" );
 
-	value = value.replace(/[^\d\.\,]/g, '');
-	value = this.numberParser(value)
+	value = value.replace(/[^\d\.\,]/g, "");
+	value = this.numberParser(value);
 
-	// return this.currencyParser( options )( value );
+	this.currencyParser( options )( value );
 	return value;
 };
 
--- a/src/currency.js
+++ b/src/currency.js
@@ -148,7 +148,7 @@ Globalize.prototype.formatCurrency = function( value, currency, options ) {
 Globalize.parseCurrency =
 Globalize.prototype.parseCurrency = function( value, options ) {
 	validateParameterPresence( value, "value" );
-
+	
 	value = value.replace( /[^\d\.\,]/g, "" );
 	value = this.numberParser( value );
 
--- a/src/currency.js
+++ b/src/currency.js
@@ -148,7 +148,14 @@ Globalize.prototype.formatCurrency = function( value, currency, options ) {
 Globalize.parseCurrency =
 Globalize.prototype.parseCurrency = function( value, options ) {
 	validateParameterPresence( value, "value" );
-	value = value.replace( /[^\d\.\,]/g, "" );
+	value = value.replace( /[^\d\.\,\-]/g, "" );
+	if ( value[0] === "-" ) {
+		value = value.replace( "-", "" );
+		value = "-".concat(value);
+	}
+	else {
+		value = value.replace( "-", "" );
+	}
 	value = this.numberParser( value );
 
 	this.currencyParser( options )( value );
--- a/src/currency.js
+++ b/src/currency.js
@@ -94,17 +94,27 @@ Globalize.prototype.currencyFormatter = function( currency, options ) {
 /**
  * .currencyParser( currency [, options] )
  *
- * @currency [String] 3-letter currency code as defined by ISO 4217.
- *
  * @options [Object] see currencyFormatter.
  *
  * Return the currency parser according to the given options and the default/instance locale.
  */
 Globalize.currencyParser =
-Globalize.prototype.currencyParser = function( /* currency, options */ ) {
+Globalize.prototype.currencyParser = function( options ) {
+	var args, numberParser, properties, returnFn;
 
-	// TODO implement parser.
+	validateParameterTypePlainObject( options, "options" );
 
+	options = options || {};
+
+	args = [ options ];
+
+	numberParser = this.numberParser( options );
+
+	// value.replace(/[^\d\.\,]/g, '');
+	// returnFn = numberParser(value)
+
+	runtimeBind( args, returnFn, [ numberParser, properties ] );
+	return returnFn
 };
 
 /**
@@ -131,15 +141,22 @@ Globalize.prototype.formatCurrency = function( value, currency, options ) {
  *
  * @value [String]
  *
- * @currency [String] 3-letter currency code as defined by ISO 4217.
- *
  * @options [Object]: See currencyFormatter.
  *
  * Return the parsed currency or NaN when value is invalid.
  */
 Globalize.parseCurrency =
-Globalize.prototype.parseCurrency = function( /* value, currency, options */ ) {
+Globalize.prototype.parseCurrency = function( value, options ) {
+	validateParameterPresence( value, "value" );
+
+	value = value.replace(/[^\d\.\,]/g, '');
+	value = this.numberParser(value)
+
+	// return this.currencyParser( options )( value );
+	return value;
 };
+
+
 
 return Globalize;
 
--- a/src/currency.js
+++ b/src/currency.js
@@ -151,9 +151,8 @@ Globalize.prototype.parseCurrency = function( value, options ) {
 	value = value.replace( /[^\d\.\,\-]/g, "" );
 	if ( value[0] === "-" ) {
 		value = value.replace( "-", "" );
-		value = "-".concat(value);
-	}
-	else {
+		value = "-".concat( value );
+	} else {
 		value = value.replace( "-", "" );
 	}
 	value = this.numberParser( value );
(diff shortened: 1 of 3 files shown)


goatandsheep:
I don't expect this build to pass as there are some code issues. I want some help working on the PR.

goatandsheep:
I can confirm that removing those symbols works very well as I tested it using a custom number parser locally. Please consider pushing this.

goatandsheep:
I need this for work. I have a workaround until this is merged, but it is not robust for all currencies.

jzaefferer:
Hey @goatandsheep thanks for the contribution! I think @rxaviers has lots on his plate and it might still be a while until he can review this.

What I can contribute: In order to get this merged, it needs to at least pass existing tests, including linters. You can run those locally, or click on the Details link next to the failed Travis build here.

For a change like this there's also additional tests required. The project has both unit and functional tests. Usually looking at the existing tests for the currency and number modules should give you an idea how to add tests for this feature.

goatandsheep:
This raises a number of questions: 

1) How do I use `currency-runtime.js` ?

2) Started testing locally and `npm test` is coming up with date issues, despite me not touching the date module. I think this must be from a previous commit.

```
>> Date Parse - should parse date according to passed timeZone in various datetime patterns
>> Message: null
>> Actual: "Sun Mar 12 2017 07:00:00 GMT-0400 (Eastern Daylight Time)"
>> Expected: "Sun Mar 12 2017 06:00:00 GMT-0400 (Eastern Daylight Time)"
>> dateParseWithTimezone@http://localhost:9001/unit/date/parse.js:46:16
>> http://localhost:9001/unit/date/parse.js:1219:30
>> run@http://localhost:9001/external/qunit/qunit/qunit.js:895:32
>> http://localhost:9001/external/qunit/qunit/qunit.js:1024:14
>> process@http://localhost:9001/external/qunit/qunit/qunit.js:583:24
>> next@http://localhost:9001/external/qunit/qunit/qunit.js:570:10
```

3) `test/compiler/cases/number.js` doesn't test `numberParser`. `test/compiler/cases/date.js` doesn't test `dateParser`. `test/compiler/cases/unit.js` doesn't test `unitParser`. It seems like there isn't really a standard on how to test parsers, so I'm not going to make one there. I can however make functional unit tests.

goatandsheep:
I'm struggling with handling the negatives, though. Coming up next!

goatandsheep:
It seems that fixed most of the lint issues, but it is complaining that not all my commits were prefixed with `Currency: `. Can we force this?

rxaviers:
Hi @goatandsheep sorry for the delay on my review. Thanks for your interest to implement this very welcome feature. From your code, it seems you're simply doing a number parser, i.e., the currency is simply being ignored. It's important that currency parser is given which currency the input is expected to be in and if the input matches that, e.g., parsing "0,05 €" should fail if you expect USD input.

src/currency.js
```diff
@@ -131,14 +141,25 @@ Globalize.prototype.formatCurrency = function( value, currency, options ) {
...
-Globalize.prototype.parseCurrency = function( /* value, currency, options */ ) {
+Globalize.prototype.parseCurrency = function( value, options ) {
+	validateParameterPresence( value, "value" );
+	value = value.replace( /[^\d\.\,\-]/g, "" );
+	if ( value[0] === "-" ) {
+		value = value.replace( "-", "" );
+		value = "-".concat( value );
+	} else {
+		value = value.replace( "-", "" );
+	}
```
rxaviers: The above digits and minus sign should be taken care by numberParser below. PS: Note that `/\d/` doesn't work on other numbering systems, e.g., arabic. It's not necessary to check for that anyway (numberParser will take care)

test/functional/currency/currency-parser.js
```diff
@@ -0,0 +1,146 @@
...
+
+QUnit.test( "should return a currency formatter", function( assert ) {
+	var de, zh;
+
+	extraSetup();
+
+	de = Globalize( "de" );
+	zh = Globalize( "zh" );
+
+	assert.equal( Globalize.currencyParser()( "$69,900.00" ), teslaS );
```
rxaviers: Parser should know what's the expected currency and assert that. We need a test where a different currency is appropriately not parsed (e.g., parser could return `null`).

test/functional/currency/currency-parser.js
```diff
@@ -0,0 +1,146 @@
...
+
+	assert.equal( Globalize.currencyParser()( "69,900.00 USD" ), teslaS );
+	assert.equal( de.currencyParser()( "69.900,00 USD" ), teslaS );
+	assert.equal( zh.currencyParser()( "69,900.00USD" ), teslaS );
+
+	assert.equal( Globalize.currencyParser()( "69,900.00 US dollars" ), teslaS );
+	assert.equal( de.currencyParser()( "69.900,00 US-Dollar" ), teslaS );
+	assert.equal( zh.currencyParser()( "69,900.00美元" ), teslaS );
+
+	assert.equal( Globalize.currencyParser()( "($1.00)" ), -1 );
```
rxaviers: We need to include a test using a different numbering system, e.g., arabic.

goatandsheep:
@rxaviers thank you so much for reviewing my PR. I'll get to the fixes, but first I wanted to address your concern about the simplicity of the function.

Perhaps I'm wrong about this opinion, but I believe the `.` vs `,` issue is related to locale, which your NumberParser handles beautifully.

Regarding currency symbols, unless it includes a value factor like the M in $10M, non-digits are simply identifiers of the type of currency, which can be already recognized by the currency code. We can even constraint the input to not allow value factors.

test/functional/currency/currency-parser.js
goatandsheep: Should this reflect in NumberParser too?

src/currency.js
goatandsheep: I actually *didn't* know `\d` won't work with non-ASCII numbers. How do I obtain the regex for those digits? Regarding this section in particular, I guess I wanted to make sure I didn't leave any `-`, but I'm realizing now that if there's a dash in the middle, it should return `NaN`. I'll take that out.

test/functional/currency/currency-parser.js
goatandsheep: Yeah I think my parser is not strict enough to pick up `NaN`

src/currency.js
goatandsheep: I can use `/\p{Decimal_Number}+/u` instead, but that requires ES2018

test/functional/currency/currency-parser.js
rxaviers: I mean  this:
```js
assert.equal( Globalize.currencyParser("USD")( "$69,900.00" ), teslaS );
// > 69900

assert.equal( Globalize.currencyParser("EUR")( "$69,900.00" ), teslaS );
// > null
```

I said `null`, but it could be `NaN` instead (like numberParser)

test/functional/currency/currency-parser.js
rxaviers: Sorry, I didn't follow your question

rxaviers:
Let's define the goal of currency parser.

It should:
- be able to parse strings created by formatCurrency; E.g., `currencyParser("USD")(currencyFormatter("USD")(10)) // 10`;

It should NOT:

- try to guess on which currency the input is at. E.g., ~~`parse("$10") // {currency: "USD", value: 10}`~~ This is error prone because figuring out currency is non deterministic (there are various currencies that use the same symbol)

In terms of implementation:
- It should not duplicate number code; it should leverage numberParser and only do the extra logic that is needed to get currency parsing, which is to check for the currency string. See how currencyFormatter works and read http://www.unicode.org/reports/tr35/tr35-numbers.html#Currencies

goatandsheep:
I agree with the goals that you've stated, but I don't agree that we need to look at the currency code to identify quantity. We don't need to guess because we don't even need to know the type of currency. All we need to do is strip away the currency symbols and parse the number. As I said earlier, `0,05 €` is not a format that would be typed in the locale `en-us`, just as `€0.05` would not be typed in the locale `fr`. `NumberParser` is conscious of locale, so as you mentioned, it is a great tool to distinguish the `,` vs `.`

rxaviers:
Sorry, if one wants to simply strip away the currency symbol and parse the number, they can use today (a) [this regexp](https://stackoverflow.com/a/27175364/798133) to strip away currency and (b) number parser to normalize the amount. This is a BAD solution for obvious reasons (i.e., not good to assume a wrong currency).

The best way to capture such information is to provide a good UI where user can select the currency and type the amount.

The currency parser, by doing what's defined above https://github.com/globalizejs/globalize/pull/795#issuecomment-372307858), would be useful to parse a controlled input (either using mask, or a future format currency to parts).

src/currency.js
frankrobert-ls: If you need something more dynamic and that's supported cross-browser you could use `new RegExp()`

frankrobert-ls:
@rxaviers @goatandsheep Are there any updates on this?

goatandsheep:
@frankrobert-ls unfortunately not because of our disagreement of opinions. @rxaviers believes that you need to extract further localization information from the currency symbol and I believe that you just need the localization information that the browser gives you.

My recommendation is just to use [this other currency parser](https://github.com/mktj/parsecurrency).

goatandsheep:
actually @rxaviers I don't mind redoing the whole PR to match the style of the above mentioned package, which does the currency matching

src/currency.js
goatandsheep: can you show me where I would add `new RegExp()`?

rxaviers:
FWIW, I am not accepting this PR because it lacks (at least) functionality that prevents adding these two unit tests.
- https://github.com/globalizejs/globalize/pull/795#discussion_r173446337
- https://github.com/globalizejs/globalize/pull/795#discussion_r173446454

I am open to accept this PR (or a subsequent one leveraging the changes made here) if the above is met.

test/functional/currency/currency-parser.js
goatandsheep: if this is dependent on `NumberParser`, should `NumberParser` also have tests in different numbering systems?

goatandsheep:
Sounds good! Will do.

test/functional/currency/currency-parser.js
rxaviers: Thanks for the clarification. Yes, it does in unit tests. Example https://github.com/globalizejs/globalize/blob/daf1651b5b6ccfa9e62b4de025c61d116431ffb6/test/unit/number/parse.js#L121

PR #806, opened 2018-03-09
Fix leaking handler in date format setup

SlexAxton:
A coworker of mine patched this locally because we were experiencing progressive slow-downs the more we rendered certain pages with lots of dates.

> They are adding and removing an event listener from cldr when we pass in options.timeZone, but since they are constructing the listener with a generator function, they aren't actually removing the one they added. I'll see if I can modify this function and check if that fixes it.

<3z,
Alex


rxaviers:
Thanks! ❤️

rxaviers:
Published in https://github.com/globalizejs/globalize/releases/tag/1.4.0-alpha.1

Issue #821, opened 2018-04-11
Rounding with significant digits precision gives wrong results


shivijais:
```
Globalize.numberFormatter({
        compact: 'short',
        minimumSignificantDigits: 1,
        maximumSignificantDigits: 3,
    })(12849872883)
```
Actual => `12.9B`
Expected => `12.8B`

Root cause for this is in the method `numberToPrecision` which first changes the precision from actual precision to a +2. and then rounds the number based on the given precision. So the number in this case `12849872883` changes to `12.850` and when finally rounded becomes `12.9`

rxaviers:
Thanks for reporting and for the initial investigation. Fix would be welcome :). Thanks

PR #830, opened 2018-06-07
Number: Compact formatting with significat digits fix

shivijais:
Fixes: #821 

Removed the logic of rounding the number twice - with a precision of {actual precision required + 2} and then again with the actual precision. This was causing wrong rounding off a number. For eg. 

```
Globalize.numberFormatter({
        minimumSignificantDigits: 1,
        maximumSignificantDigits: 3,
    })(12.849872883)
```

Here the number 12.849872883 is first converted to 12.850 and then when finally rounded off becomes 12.9. But Ideally this should be formatted to 12.8.

### DIFF
--- a/test/unit/number/format.js
+++ b/test/unit/number/format.js
@@ -462,6 +462,18 @@ QUnit.test( "numbers should support significant digits in compact mode", functio
 		maximumSignificantDigits: 3,
 		minimumSignificantDigits: 1
 	})), "-127M" );
+
+	assert.equal( format( 12849872883, properties("0", en, {
+		minimumSignificantDigits: 1,
+		maximumSignificantDigits: 3,
+		compact: "short"
+	})), "12.8B" );
+	
+	assert.equal( format( 12850172883, properties("0", en, {
+		minimumSignificantDigits: 1,
+		maximumSignificantDigits: 3,
+		compact: "short"
+	})), "12.9B" );
 });
 
 QUnit.test( "numbers should support rounding in compact mode", function( assert ) {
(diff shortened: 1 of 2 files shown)


shivijais:
@rxaviers Can you please take a look at this PR. The travis-ci is failing on timeout issue with `https://github.com/SlexAxton/messageformat` which is now moved to `https://github.com/messageformat/messageformat`

test/unit/number/format.js
```diff
@@ -462,6 +462,12 @@ QUnit.test( "numbers should support significant digits in compact mode", functio
 		maximumSignificantDigits: 3,
 		minimumSignificantDigits: 1
 	})), "-127M" );
+
+	assert.equal( Globalize.formatNumber( 12849872883, {
+		minimumSignificantDigits: 1,
+		maximumSignificantDigits: 3,
+		compact: "short"
+	}), "12.8B" );
```
rxaviers: Since these are unit tests, you don't have `Globalize` available. You need to use `format` (according to the above tests).

rxaviers: Let's also add a test where `12850172883` becomes `12.9B`

rxaviers:
Thank you so far

rxaviers:
Thanks for the quick fix. Tests are passing, build is failing simply because of commit message which I can fix.

Having said that, I want to be sure the code simplification in the round function doesn't cause any side effect and therefore I plan to review this again. It's on my end, I hope to do it soon and will update this PR.

shivijais:
@rxaviers any update on this?

rxaviers:
Thank you! Released as 1.4.0-alpha.3

PR #839, opened 2018-07-27
Unit: Fix formatUnit to display formatted number with compound units

### LINKED ISSUE #838: formatUnit does not display the formatted number when using a compound unit pattern
When using a unitFormatter that takes a custom numberFormatter as an option, the formatter is called but the output is not used.
The number is dumped directly into the returned message instead of using the formatted value.

The following example illustrates an example.
```js
var customNumberFormatter = Globalize('fr').numberFormatter({ maximumFractionDigits: 1 });
var metersPerLiter = formatUnit(1.234567, 'meter-per-liter', { form: 'narrow', numberFormatter: customNumberFormatter });
var milesPerHour = formatUnit(1.234567, 'mile-per-hour', { form: 'narrow', numberFormatter: customNumberFormatter });

assert(metersPerLiter).toBe('1,2m/l');  // Fails as metersPerLiter === '1.234567m/l' - note the period (.) instead of comma (,) too.
assert(milesPerHour).toBe('1,2mph'); // True
```

EdwardSalter:
When using compound units with the `formatUnit` function, it now correctly displays the formatted value and instead of dumping the raw value into the message.

### DIFF
--- a/src/unit/format.js
+++ b/src/unit/format.js
@@ -36,7 +36,7 @@ return function( value, numberFormatter, pluralGenerator, unitProperties ) {
 		dividendProperties = unitProperties[ 0 ];
 		divisorProperties = unitProperties[ 1 ];
 
-		dividend = formatMessage( dividendProperties[ pluralValue ], [ value ] );
+		dividend = formatMessage( dividendProperties[ pluralValue ], [ formattedValue ] );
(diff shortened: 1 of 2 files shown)


EdwardSalter:
I have now signed the CLA.

EdwardSalter:
Fixes #838

rxaviers:
Good catch! Thanks for your fix!

PR #855, opened 2019-02-27
Fix strict mode

### LINKED ISSUE #854: Number formatting fails in strict mode
When bundling for ES6 modules, the scripts are by default loaded in 'strict mode'. The number formatting in globalize seems to be violating strict mode in these lines:

https://github.com/globalizejs/globalize/blob/master/src/number/format-properties.js#L74

```
properties[ 2 ] = negativeSuffix[ 2 ] = 1; // minimumIntegerDigits
properties[ 3 ] = negativeSuffix[ 3 ] = 0; // minimumFractionDigits
properties[ 4 ] = negativeSuffix[ 4 ] = 0; // maximumFractionDigits
```

It fails with `TypeError: Cannot create property '2' on string ''`

ashubham:
Adding properties to strings is disallowed in strict mode. Fixes #854

### DIFF
--- a/src/number/format-properties.js
+++ b/src/number/format-properties.js
@@ -71,11 +71,11 @@ return function( pattern, cldr, options ) {
 		// the value to be formatted, though we're always using 1 as a simplification, because the
 		// number won't be zero-padded since we chose the right format based on the scale, i.e.,
 		// we'd never see something like `003M` anyway.
-		properties[ 2 ] = negativeSuffix[ 2 ] = 1; // minimumIntegerDigits
-		properties[ 3 ] = negativeSuffix[ 3 ] = 0; // minimumFractionDigits
-		properties[ 4 ] = negativeSuffix[ 4 ] = 0; // maximumFractionDigits
-		properties[ 5 ] = negativeSuffix[ 5 ] = // minimumSignificantDigits &
-			properties[ 6 ] = negativeSuffix[ 6 ] = undefined ; // maximumSignificantDigits
+		properties[ 2 ] = 1; // minimumIntegerDigits
+		properties[ 3 ] = 0; // minimumFractionDigits
+		properties[ 4 ] = 0; // maximumFractionDigits
+		properties[ 5 ] = // minimumSignificantDigits &
+			properties[ 6 ] = undefined ; // maximumSignificantDigits
 
 		properties[20] = numberCompact( options.compact, cldr );
 	}


src/number/format-properties.js
```diff
@@ -71,11 +71,11 @@ return function( pattern, cldr, options ) {
...
-		properties[ 2 ] = negativeSuffix[ 2 ] = 1; // minimumIntegerDigits
-		properties[ 3 ] = negativeSuffix[ 3 ] = 0; // minimumFractionDigits
-		properties[ 4 ] = negativeSuffix[ 4 ] = 0; // maximumFractionDigits
-		properties[ 5 ] = negativeSuffix[ 5 ] = // minimumSignificantDigits &
-			properties[ 6 ] = negativeSuffix[ 6 ] = undefined ; // maximumSignificantDigits
+		properties[ 2 ] = 1; // minimumIntegerDigits
+		properties[ 3 ] = 0; // minimumFractionDigits
+		properties[ 4 ] = 0; // maximumFractionDigits
+		properties[ 5 ] = // minimumSignificantDigits &
+			properties[ 6 ] = undefined ; // maximumSignificantDigits
```
rxaviers: Please, what about setting the `negativeSuffix` items?

rxaviers: Please, can you explain how this change fixes the problem? Thanks

rxaviers:
Thanks for your PR, I left some comments.

src/number/format-properties.js
ashubham: negativeSuffix is a string.

Setting `negativeSuffix[idx] = <value>` is not allowed in `strict mode` of javascript. You can try and reproduce it locally with:
```
(function strictMode() {
`use strict`;
let str = "";
str[2] = 1;
})()
```

ashubham: > Please, what about setting the `negativeSuffix` items?

I could not find a reason to set these to negativeSuffix in the first place. It will be great if you have some pointers towards that, so that I might come up with a workaround, if needed.

ashubham:
@rxaviers Thanks for taking a look. Responded.

src/number/format-properties.js
rxaviers: That's true!! What I found out: I believe `negativeSuffix` was a typo. It should have been `negativeProperties`. Although, it also doesn't make sense to set `negativeProperties` at all. This variable is only used as an auxiliary variable to get `negativePrefix` and `negativeSuffix`. Therefore, I believe that's was mistake when compact support was implemented. That mistake doesn't cause bug, because this is variable isn't used and it's ignored.

Your change looks good to me.

rxaviers:
Thank you!

Issue #706, opened 2017-03-31
formatNumber returns NaN for 0, if max/minimumSignificantDigits is set


AndreR:
Hello,

If you call the following function, it returns "NaN %":

`Globalize.formatNumber(0, {style:'percent',minimumSignificantDigits: 1,maximumSignificantDigits: 3})`

If you leave out the significant digits settings, it returns an empty string, as I would expect.

rxaviers:
Thanks for reporting this bug.

PR #865, opened 2019-04-16
Number: added zero check for toPrecision

hmizutanitsi:
Component: number

formatNumber returns NaN for 0, if max/minimumSignificantDigits is set

Fixed: #706


hmizutanitsi:
don't know how to change commit message. Redoing PR

PR #899, opened 2020-05-20
Fix dayPeriod and timeZoneName in documentation

dantman:
Globalize actually returns lower case `dayperiod`. But the documentation says `dayPeriod`.

This is confusing because `Intl.DateTimeFormat`'s `formatToParts` does return `dayPeriod` but Globalize diverges from this and returns lower case `dayperiod`. This mistake in the documentation can lead to tricky bugs where a developer writes code according to the documentation checking for `dayPeriod` which doesn't work.


dantman:
Additionally Globalize actually returns a `zone` part with zone info, not a `timeZoneName` part.

rxaviers:
Thank you

PR #904, opened 2020-08-06
Bump CLDRjs to 0.5.4

morwoen:
Latest version contains multiple fixes and improvements which would be good to have in Globalize :)

Signed-off-by: Georgi Tsaklev <georgi.tsaklev@skyscanner.net>


robaw:
Hi @rxaviers here is the bump in version for cldrJS with our recent fix in it.

Issue #811, opened 2018-03-09
Stop supporting bower in a version 2.0?


rxaviers:
We used bower in the days where npm didn't have flatten dependencies support. This has changed a long time and people has moved towards npm for package management for some time by now.

The goal of this issue is to collect community feedback (if any) about moving this path forward.

jzaefferer:
Its time, as bower itself tells users: https://bower.io/blog/2017/how-to-migrate-away-from-bower/

rxaviers:
Apparently closed by #926

stukalin:
@rxaviers I haven't really fully dropped bower, just one package. So this should stay opened for a bit

rxaviers:
Thanks

PR #926, opened 2021-11-08
cldr-data is now loaded from package.json

### LINKED ISSUE #544: CLDR dayPeriods
CLDR data provides a variety of day periods:

``` js
new Cldr( "en" ).main([ "dates/calendars/gregorian/dayPeriods", "format", "abbreviated" ]);
/* outputs: 
{
  afternoon1: "in the afternoon",
  am: "AM",
  am-alt-variant: "am",
  evening1: "in the evening",
  midnight: "midnight",
  morning1: "in the morning",
  night1: "at night",
  noon: "noon",
  pm: "PM",
  pm-alt-variant: "pm"
}
*/
```

The `skeleton` option can be used to format the time, but it always includes either the `am` or `pm` settings since most formats use "a". In the spec you can use "B" in the skeleton, but it only uses `night1` and possibly `morning1`? "b" is also allowed to set "noon" and "midnight" specific times.

I would like to propose adding a `dayPeriod` option which would allow use of `C` ([per spec](http://www.unicode.org/reports/tr35/tr35-dates.html#Date_Field_Symbol_Table)); additional options maybe be needed to further customize the day period string used, maybe something similar to this:

``` js
var opts = {
  skeleton : "hm",
  dayPeriod : "C", // "a", "b", "B" or "C" (per spec) for custom day periods
  // the below option only works with "B" or "C"  set in dayPeriod
  setPeriods : {
    "00:00" : "midnight",
    "00:01" : "night1",
    "06:00" : "morning1",
    "12:00" : "noon",
    "12:01" : "afternoon1",
    "18:00" : "evening1",
    "21:00" : "night1"
  }
};
Globalize( "en" ).dateFormatter( opts )( new Date( 2010, 10, 1, 17, 55 ) );
// outputs: "5:55 in the afternoon"
Globalize( "en" ).dateFormatter( opts )( new Date( 2010, 10, 1, 19, 55 ) );
// outputs: "7:55 in the evening"
```

stukalin:
Addresses #811 and moves the `cldr-data` package to npm. The latest version of the package is 36 whether the latest version of cldr is 40 now, but i guess it's now closer rather than 25 :D

Anyways I did this one in order to obtain the dayPeriods data which is required to resolve #544.

### DIFF
--- a/test/unit/number/format-properties.js
+++ b/test/unit/number/format-properties.js
@@ -66,6 +66,7 @@ QUnit.test( "should return nanSymbol", function( assert ) {
 });
 
 QUnit.test( "should return symbolMap", function( assert ) {
+	// debugger;
 	assert.deepEqual( properties( "0", es )[ 18 ], {
 		"%": "%",
 		"+": "+",
@@ -77,11 +78,11 @@ QUnit.test( "should return symbolMap", function( assert ) {
 	});
 
 	assert.deepEqual( properties( "0", fa )[ 18 ], {
-		"%": "٪",
-		"+": "\u200e+\u200e",
-		",": "٬",
-		"-": "\u200e−",
 		".": "٫",
+		",": "٬",
+		"%": "٪",
+		"+": "‎+",
+		"-": "‎−",
 		"E": "×۱۰^",
 		"‰": "؉"
 	});
--- a/.bowerrc
+++ b/.bowerrc
@@ -1,6 +1,3 @@
 {
-  "directory": "external",
-  "scripts": {
-    "postinstall": "node ./node_modules/cldr-data-downloader/bin/download.js -i external/cldr-data/index.json -o external/cldr-data/"
-  }
+  "directory": "external"
 }
--- a/bower.json
+++ b/bower.json
@@ -11,7 +11,6 @@
     "cldrjs": "^0.5.4"
   },
   "devDependencies": {
-    "cldr-data": ">=25",
     "es5-shim": "3.4.0",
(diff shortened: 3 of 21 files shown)


stukalin:
@rxaviers may I ping you on this one? I'd like to do a further contribution but it waits till this one will be merged.

stukalin:
@rxaviers just another friendly ping

rxaviers:
Thank you!

test/unit/number/format-properties.js
```diff
@@ -77,11 +78,11 @@ QUnit.test( "should return symbolMap", function( assert ) {
...
 	assert.deepEqual( properties( "0", fa )[ 18 ], {
-		"%": "٪",
-		"+": "\u200e+\u200e",
-		",": "٬",
-		"-": "\u200e−",
 		".": "٫",
+		",": "٬",
+		"%": "٪",
+		"+": "‎+",
+		"-": "‎−",
```
rxaviers, line 85: I merged the PR, but realized this test change is using the invisible character itself instead of the explicit `\u200e`. Can we revert that in a new PR? Thanks

rxaviers: The explicit character makes reviewing (and understanding) the test easier.

stukalin: @rxaviers sure! I'll create a new PR to fix the issue #544 and I'll rever it there

rxaviers: Thank you 🙇

PR #952, opened 2025-09-26
Add support for literals in compact patterns

### LINKED ISSUE #917: Literal quotes not removed from compact number/currency formats
Hey :wave: 

As you can see in the following example, literal quotes are not fully removed from compact number/currency formats.  

```js
Globalize("de").formatNumber(1_000_000, { compact: "short" });
// => 1 Mio'.'

Globalize("de").formatCurrency(1_000_000, "EUR", { compact: "short" });
// => 1 Mio'.' €
```

This is a regression that was introduced in [1.5.0](https://github.com/globalizejs/globalize/releases/tag/1.5.0) with this commit [8b2a006](
https://github.com/globalizejs/globalize/commit/8b2a006d759e83f2f8fb898d36fe033db9edb3c3#diff-74c9f6f524f9fc9829dba4c29952bcd7d08b282664142f13b9c35bdbc9e1a396L143).

What would be a good approach to fix this? Does it make sense to remove literal quotes specifically from the `compact` part, or do we need to take more parts into consideration? Literal quotes are already partially removed in `stringToParts`.

### LINKED ISSUE #946: Globalize date formatter prints quotes around string literals when using short formats
When running the following
```
var Globalize = require( "globalize" );
Globalize.load( require( "cldr-data" ).entireSupplemental() );
Globalize.load( require( "cldr-data" ).entireMainFor( "en", "es", "bg" ) );
console.log("Globalize curr with formatCurrency:" + Globalize("bg").formatCurrency(11231241223, "BGN", {
   compact: "short"
}));

console.log("Globalize curr with currencyFormatter:" + Globalize("bg").currencyFormatter("BGN", {
     compact: "short"
})(11231241223)); 
```

we get:
```
Globalize curr with formatCurrency:11 млрд'.' лв.
Globalize curr with currencyFormatter:11 млрд'.' лв.
```
I believe this is incorrect, and that the single quotes should not be there

The data is coming, I believe, from [here](https://github.com/unicode-org/cldr-json/blob/main/cldr-json/cldr-numbers-modern/main/bg/numbers.json)

movermeyer-stripe:
### What are you trying to accomplish?

The code was not accounting for the possibility of string literals in compact patterns ([example](https://github.com/unicode-org/cldr/blob/ae7926603d478433266d79b88a8d05d30b7da3cf/common/main/de.xml#L6629)).

This was causing issues with the compact rendering of numbers in some languages.  

### What approach did you choose and why?

Copied the same regex for detecting literals in the `stringToParts` function, and applied it to `compactPattern`.

### What should reviewers focus on?

That the regex makes sense.

It's safe to add the `'` to the negated character class, since a `'` is either the start of a string literal or should be escaped with a second `'`, both of which are covered by the first case in the alternation.  

### The impact of these changes

Fixes https://github.com/globalizejs/globalize/issues/917
Fixes https://github.com/globalizejs/globalize/issues/946

### DIFF
--- a/src/number/format.js
+++ b/src/number/format.js
@@ -174,11 +174,11 @@ return function( number, properties, pluralGenerator ) {
 		// format-properties.
 		aux = function( string ) {
 			var parts = [];
-			string.replace( /(\s+)|([^\s0]+)/g, function( _garbage, space, compact ) {
+			string.replace( /('(?:[^']|'')+'|''|\s+)|([^\s'0]+)/g, function( _garbage, literal, compact ) {
 
 				// Literals
-				if ( space ) {
-					partsPush( parts, "literal", space );
+				if ( literal ) {
+					partsPush( parts, "literal", removeLiteralQuotes( literal ) );
 					return;
 				}
 
--- a/test/functional/currency/format-currency.js
+++ b/test/functional/currency/format-currency.js
@@ -1,5 +1,7 @@
 define([
 	"globalize",
+	"json!cldr-data/main/de/currencies.json",
+	"json!cldr-data/main/de/numbers.json",
 	"json!cldr-data/main/en/currencies.json",
 	"json!cldr-data/main/en/numbers.json",
 	"json!cldr-data/supplemental/currencyData.json",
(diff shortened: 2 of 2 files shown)


rxaviers:
Released 1.7.1

